PR #5, opened 2021-01-28
Adding dm2 bis

rgalicher:


rgalicher:
Remove code from Main_EFC
Small modification in coronagraph structure (sorry Johan)

PR #6, opened 2021-02-01
Adding dm2 bis

rgalicher:


rgalicher:
Major modifications everywhere. Stable version

PR #18, opened 2021-03-17
Integrate new mft function

johanmazoyer:
We modify the mft to allow the easy offset in both direction (input array and output array).

The MFT is now more complete (can be ery different in each direction, etc) but if used with only one parameter for dim etc, it automatically do both direction


johanmazoyer:
2 very small comflicts resolved in the setup (I removed pypng that was useless) and a in a comment in instrumentsimu_functions.py

PR #19, opened 2021-03-18
Labview compatibility and PW is repared

rgalicher:


johanmazoyer (approved):
ok this is working in PW. Default parameter file is for 1 DM

PR #22, opened 2021-03-20
Small modif r ga

rgalicher:
Correction of small things (name of estimation is now converted to lowercase, and so on)

### DIFF
--- a/Asterix/Main_EFC_THD.py
+++ b/Asterix/Main_EFC_THD.py
@@ -187,7 +187,7 @@ def create_interaction_matrices(parameter_file,
             vectoressai, showsvd = wsc.createvectorprobes(
                 thd2, amplitudePW, posprobes, dim_sampl, cut,
                 thd2.wavelength_0)
-        fits.writeto(intermatrix_dir + filePW + ".fits", vectoressai)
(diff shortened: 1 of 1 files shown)


Asterix/Main_EFC_THD.py
```diff
@@ -942,6 +937,14 @@ def correctionLoop(parameter_file,
 
     ## SAVING...
     header = useful.from_param_to_header(config)
+    if thd2.DM1.active == True:
+        cut_phaseDM1 = np.zeros((nbiter + 1, 2 * thd2.prad, 2 * thd2.prad))
+        for it in np.arange(nbiter + 1):
+            cut_phaseDM1[it] = proc.crop_or_pad_image(phaseDM1[it],
```
johanmazoyer, line 943: why do you absolutely want to crop them to the pupil plane ? they are not that big (overpad pupil is only 25% larger than pupil). More importantly, for the DM that is outside of the pupil plane, the movement of the actuator outside the pupil is relevant because the beam is larger than the pupil. why throw away this information to save a few Mbytes ?

johanmazoyer (approved):

PR #29, opened 2021-04-15
Clean correc bis

johanmazoyer:
a lot of cleaning of the correction class
the basis is now done at a single position (there fore which in pup in now only use inside the basis) and can be replaced easily by others when needed.

not sure mis registration is currently workin well!

### DIFF
--- a/Asterix/Main_EFC_THD.py
+++ b/Asterix/Main_EFC_THD.py
@@ -448,7 +448,7 @@ def correctionLoop(parameter_file,
                 #### to be removed #### #### #### #### #### #### #### #### ####
 
                 if thd2.DM1.active == True:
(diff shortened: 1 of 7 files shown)


Asterix/Main_EFC_THD.py
```diff
@@ -294,41 +299,25 @@ def correctionLoop(parameter_file,
...
-
     #initalize the corrector
     correc = Corrector(Correctionconfig,
                        thd2,
                        mask_dh,
                        estim,
-                       matrix_dir=intermatrix_dir)
-
-
-    ## Adding error on the DM model. Now that the matrix is measured, we can
```
rgalicher, line 316: where can we add the misregistration?
Is it done somewhere else?

rgalicher (approved):

PR #50, opened 2022-09-12
Add conda environment file

ivalaginja:
Up to this point, the installation instructions in the PDF documentation lists manually how to create the conda environment for Asterix. In this PR, I added an `environment.yml` file that can be run to do that without having to manually specify the conda dependencies. I also removed the duplicated dependencies from `setup.py` so that there are no double installations between conda and pip.

I also updated the readme to have quick instructions for the installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,26 @@
-# Package Asterix
+# Asterix
 
-Asterix is a package to simulate high contrast-correction algorithns with various coronagraphs, developped at LESIA / Paris Observatory.
+Asterix is a package to simulate high-contrast imaging correction algorithms with various coronagraphs, developed at LESIA / Paris Observatory.
 
 Please find the latest documentation (pdf file) here : https://www.dropbox.com/s/wjiwci2r6yzudvj/asterix.pdf?dl=1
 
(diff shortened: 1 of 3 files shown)


johanmazoyer:
hi, I'm not sure I want to take the requirements out of the setup. What happens if someone installs Asterix from setup in a different environment which does not contains the right packages, now that the requirements are not in the setup ? 

I sometimes install Asterix in another environment, usually when I need it combined with another one (e.g. Asterix + pyklip for simulation and data reduction, or Asterix + hcipy to comapre both). Creating a stand alone env and install the requirements with conda is the preferred solution for the users but not the required solution. 

Finally, if it is in the setup there are no "double installations between conda and pip". The pip install with setup just acknowledge that all requirements are satisfied and just install Asterix

ivalaginja:
Fair points about the requirements in the `setup.py`. I will iterate on this.

johanmazoyer:
also careful, they'll be conflicts with the current most advances branch (new_propagation_functions) because I modified the setup.py because I had removed most of the skimage use and therefore was able to be less constraining about skimage version and there fore python version. I think these modifications should be done and PR from the new_propagation_functions branch

johanmazoyer:
yes, better. it's still written 3.10 in the setup.py

johanmazoyer (approved):

PR #51, opened 2022-09-12
Fix label bug in example script

ivalaginja:
While running the example script `Example_Run_Asterix.py`, the four WFS&C example in there threw an error because the lable variable wasn't formatted appropriately. This PR fixes that and the example script now finished without errors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 #ignore file extension
 *.pyc
 *.png
+*.ipynb_checkpoints
 
 # to avoid that people sync their own param file in the git dir
 *.ini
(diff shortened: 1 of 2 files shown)


johanmazoyer:
not sure what are .ipynb_checkpoints but it's probably useful  for nb Btw if you want to create a nb from the tutorial, it would probably be a lot more easy to run and understand so be my guess

johanmazoyer (approved):

ivalaginja:
The `.ipynb_checkpoints` are checkppoint files that save your manual savepoints from a notebook. If a notebook is in a git repository though, git will take care of saving "checkpoints" through the individual commits, so they're not needed.

And yes, I am already working on a tutorial notebook, I hope to have a basic version done today or tomorrow 👍

PR #52, opened 2022-09-13
small modifs plots should avoid empty pop ups

johanmazoyer:
Small modifications to do better plotting of images in the loop and avoid empty windows popping in some os


johanmazoyer:
Salut @rgalicher , tu peux me dire si ca regle ton soucis de pop ups vides sur ton ordi ?

PR #53, opened 2022-09-14
Add first pytest

ivalaginja:
I am adding a first pytest to set up an example in case there are other tests in the future that need to find a place. Currently no tests are deployed automatically but it's handy to at least have them set up to be able to run them offline.


johanmazoyer (approved):
thanks

PR #56, opened 2022-09-16
update readme and install section in pdf manual

johanmazoyer:
Edited by @ivalaginja : Always good to provide a basic description with a PR! :)

Deprecation of `setuptools` command line functionality was introduced here: https://github.com/pypa/setuptools/pull/2824
And a summary of what is happening is given here: https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

PR #57, opened 2022-09-17
New propagation functions

johanmazoyer:
Time to merge this one, these was important modifications but has moved already far from the master and need to be merged. Main changes
-> added Mehdi's function called fft_choosecenter an fft with options to choose the centering
-> got rid of almost all centering and rescaling with skimage
-> added rebin for pupils : Non-clear pupils stored in .fits can now only be used with specific number of pixel in pup fixed number that has to be a divisor of 500 (size of the pup in the .fits) so that they are binned


ivalaginja:
@johanmazoyer there is a conflict in this PR after merging #56, it looks minor but I don't dare touching it for you because it's a rather large PR overall. It might also help to rebase these kind of PRs in the future rather than merging master into them. Let me know if you need any assistance with rebasing this PR and/or resolving the conflict (looks like it's just something in the documentation)>

Asterix/Optical_System_functions.py
```diff
@@ -1806,6 +1803,8 @@ def creatingpushact(self, DMconfig):
         # do the first FT only once
         ft_actu = np.fft.fftshift(np.fft.fft2(np.fft.fftshift(actshapeinpupil), norm="ortho"))
 
+        # ft_actu = np.fft.fftshift(np.fft.fft2(np.fft.fftshift(actshapeinpupil), norm="ortho"))
+
```
ivalaginja, line 1807: This looks like it can be deleted?

ivalaginja (commented):
There is a *huge* amount of linebreak changes smuggled into this PR so it makes it really, really hard to review the actual changes this PR is about, but I think I found them all.
No major comments, it just seems like there is a commented-out line that could be deleted, otherwise it's confusing to see two FTs right after another while one is commented out.

johanmazoyer:
Yes I did a yapf format on the whole Asterix, I dot it now and then to keep the code clean. I agree that it should be done on special PR where nothing else is done to make the reviews more readable.

ivalaginja (approved):

PR #59, opened 2022-09-20
Unify class names in OpticalSystem module

ivalaginja:
Class names in Python are written in camel case [according to PEP8](https://peps.python.org/pep-0008/#class-names). Some of the classes in Asterix were already adhering to that, others no. This PR unifies this and removes some obvious comments that repeat the class names.

A summary of Python naming conventions can be found here:
https://visualgit.readthedocs.io/en/latest/pages/naming_convention.html

It seemed like a good time to do this since we just closed all PRs (except for #52 which only changes plotting) before I start working on more substantial changes.

Needs #60 to me merged first.

### DIFF
--- a/source/create_my_testbed.rst
+++ b/source/create_my_testbed.rst
@@ -8,7 +8,7 @@ Optical System
 
 Asterix have been thought from the beginning to be able to easily adapt to new configurations of the testbed 
 wihtout major changes. This modularity is based on the ``Asterix.Optical_System_functions.OpticalSystem`` class.
-An Optical_ystem is a part of the testbed which starts and ends in a pupil plane, which allows them to be easily
+An OpticalSystem is a part of the testbed which starts and ends in a pupil plane, which allows them to be easily
 concatenated. To be able to be well concatenated, they must all be using the same general parameters (physical 
 and numerical parameters). These parameters are stored in the first part of the parameter file, common to 
 all ``OpticalSystem``: [modelconfig].
--- a/source/create_my_testbed.rst
+++ b/source/create_my_testbed.rst
@@ -7,18 +7,18 @@ Optical System
 +++++++++++++++++++++++
 
 Asterix have been thought from the beginning to be able to easily adapt to new configurations of the testbed 
-wihtout major changes. This modularity is based on the ``Asterix.Optical_System_functions.Optical_System`` class.
-An Optical_System is a part of the testbed which starts and ends in a pupil plane, which allows them to be easily 
+wihtout major changes. This modularity is based on the ``Asterix.Optical_System_functions.OpticalSystem`` class.
+An Optical_ystem is a part of the testbed which starts and ends in a pupil plane, which allows them to be easily
 concatenated. To be able to be well concatenated, they must all be using the same general parameters (physical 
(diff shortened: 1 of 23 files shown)


ivalaginja:
@johanmazoyer ready to go now.

source/create_my_testbed.rst
```diff
@@ -7,18 +7,18 @@ Optical System
 +++++++++++++++++++++++
 
 Asterix have been thought from the beginning to be able to easily adapt to new configurations of the testbed 
-wihtout major changes. This modularity is based on the ``Asterix.Optical_System_functions.Optical_System`` class.
-An Optical_System is a part of the testbed which starts and ends in a pupil plane, which allows them to be easily 
+wihtout major changes. This modularity is based on the ``Asterix.Optical_System_functions.OpticalSystem`` class.
+An Optical_ystem is a part of the testbed which starts and ends in a pupil plane, which allows them to be easily
```
johanmazoyer, line 11: S

source/create_my_testbed.rst
```diff
@@ -28,14 +28,14 @@ Parameter file can be read using ``Asterix.fits_functions.read_parameter_file``
...
-    generic_os = OptSy.Optical_System(modelconfig)
+    generic_os = OptSy.OpticalSystem(modelconfig)
 
 
 For the moment, this ``generic_os`` does not do anything, it's like an empty pupil plane. 
 
-Each ``Optical_System`` has an attribute function ``EF_through`` which describes the effect this Optical_System has 
-on the electrical field when going through it. Once this function is defined, all Optical_System have access to 
-a library of functions, common to all ``Optical_System`` : ``todetector`` (electrical field in the next focal plane), 
+Each ``OpticalSystem`` has an attribute function ``EF_through`` which describes the effect this Optical System has
```
johanmazoyer, line 36: No space

source/create_my_testbed.rst
ivalaginja: Nice catch, fixed.

source/create_my_testbed.rst
ivalaginja: Are you sure? I actually think it should just be "optical system" (lower case) since it talks about the actual optical system you are simulating rather than the class object `OpticalSystem`. Thoguhts?

johanmazoyer:
Thnaks, it better, Is there others major name change we could do now, while we're at it ?

source/create_my_testbed.rst
johanmazoyer: yeah i see your point. ok we can leave it like that

ivalaginja:
Yes I have asked myself that too. Let me come back to you about that after lunch.

source/create_my_testbed.rst
ivalaginja: Ok. I still pushed the change to lower-case letters.

ivalaginja:
@johanmazoyer I recommend to merge this in the meantime, so that we have a clear cut.

johanmazoyer:
well no, because it will require people using Asterix to change their own codes, and I'd like if possible to do it only once (minimize the major changes that will break their codes). We can do a second PR and push both of them at the same time if you prefer

johanmazoyer (approved):
ok for me

PR #61, opened 2022-09-21
Function renaming

ivalaginja:
Rename some functions to make it easier to discern between functions and classes.


johanmazoyer (approved):
ok

PR #62, opened 2022-09-22
reorganize all asterix modules

johanmazoyer:
Reorganisation of the package in folders :
- loop
- optics
- optical_systems
- utils
- wfsc

Some modules have been cut in smaller modules, notably optical_systems.py in :
- optical_systems
- coronagraph
- deformable_mirror
- pupil
- testbed

and wsc_functions in :
- wf_control_functions
- wf_sensing_functions
- invert_svd

Open to comments to further improve


Asterix/loop/correction_loop.py
```diff
@@ -155,7 +146,7 @@ def correction_loop(testbed: OptSy.Testbed,
     return Resultats_correction_loop
 
 
-def correction_loop_1matrix(testbed: OptSy.Testbed,
+def correction_loop_1matrix(testbed: Testbed,
```
ivalaginja, line 149: I personally don't think we need type hints at all in Asterix (specifying the type of a function variable in the function signature). It's hard enough to enforce proper docstrings, and for all goals and purposes of Asterix, type hints add zero benefit, so I would just start dropping them altogether.

Asterix/tuto_asterix_model.py
```diff
@@ -152,7 +155,7 @@
 
 # Finally, we can initialize some DMs. DMs can be in a pupil or outside a pupil.
 # In the default configuraiton file, this one is in a pupil plane.
-DM3 = OptSy.DeformableMirror(modelconfig,
+DM3 = DeformableMirror(modelconfig,
```
ivalaginja, line 158: Don't forget to fix the indentation in cases like this.

Asterix/tuto_asterix_model.py
```diff
@@ -171,7 +174,7 @@
 #                               - A list of the same size containing the names of those systems so that you can access them
 # The list order is from the first optical system to the last in the
 # path of the light (so usually from entrance pupil to Lyot pupil).
-testbed_1DM = OptSy.Testbed([pup_round, DM3, corono],
+testbed_1DM = Testbed([pup_round, DM3, corono],
```
ivalaginja, line 177: See comment about indentation.

Asterix/utils/gaussians.py
```diff
@@ -0,0 +1,85 @@
+
+import numpy as np
+import scipy.optimize as opt
+
```
ivalaginja, line 4: Remember that PEP8 asks for two blank lines before a function definition (not class methods though)>

Asterix/wfsc/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+    'THD_quick_invert',
+]
+
+from .estimator import *
+from .corrector import *
+from .maskDH import *
+from .thd_quick_invert import *
+# from .coronagraph import *
+# from .deformable_mirror import *
+# from .testbed import *
```
ivalaginja, line 14: If unused, delete.

Asterix/wfsc/corrector.py
```diff
@@ -7,12 +7,13 @@
 import numpy as np
 from astropy.io import fits
 
-import Asterix.save_and_read as useful
-import Asterix.optical_systems as OptSy
-from Asterix.estimator import Estimator
-import Asterix.WSC_functions as wsc
-from Asterix.THD_quick_invert import THD_quick_invert
+from .estimator import Estimator
```
ivalaginja, line 11: Unnecessary blank line.
Also, I see that sometimes you use absolute imports (`from Asterix.blah import this`) and sometimes relative imports like here (`from .blah import this`). Is there a reason for this?

Asterix/wfsc/corrector.py
```diff
@@ -7,12 +7,13 @@
...
 
+from .thd_quick_invert import THD_quick_invert
+from .wf_control_functions import *
```
ivalaginja, line 13: Please don't do blank imports. It's bad practice and can get confusing.

ivalaginja (changes requested):
I think this looks pretty good, but I do have the feeling like it is a little over-done at this point. I think (?) the main goal is to group already existing but slightly concealed groups better, not necessarily to force everything that exists in Asterix into a group - it is ok to leave thins in the top-level location if there just is not consistent group (yet).

Below are my suggestions for the reorganization, all at your own discretion. I left a couple of inline comments but I will pay more attention to that in the next round once the new structuring is set.

- I recommend using this PR to rename the folder `Model` to `model`
- I very much disagree with the need for the folder `loop`. The name alone isn’t very specific and the module `correction_loop.py` can easily be moved to `wfsc`, while `save_results.py` can be moved to `utils` or even the top-level folder.
- I am not sure I agree with `[maskDH.py](http://maskDH.py)` being in the folder `wfsc`. One could need to use this to define a DH without running closed-loop WFS&C. I see that the parameters for the DH are being passed in from `Correctionconfig`, that’s the only tie to wfsc here. Even so, I would probably rather put it in the top-level package since there is simply nothing comparable to this right now.
- I also find the separation between the fodlers `optics` and `optical_systems`a little arbitrary. The only two modules that are currently in `optics` have to do with aberrations or with propagations. I would almost say to just resolve this folder and put these two modules back into the top-level package. If there is ever a new module to be made that touch aberrations or propagations, we can always regroup then. And this would also leave you the name.
- If you do the above point, you could call `optics` what is now called `optical_systems`. Always better to have single-word folders.

Asterix/loop/correction_loop.py
johanmazoyer: I disagree, I started using that because if you indicate those, vscode allow me to use autocomplete with all the testbed methods, it's super useful :-) 
For int, or floor I agree it's useless, and I don't indicate it, but for complex classes, its awesome !

Asterix/tuto_asterix_model.py
johanmazoyer: we agreed to do yapf format in a separate PR so that it's easier to read

Asterix/tuto_asterix_model.py
johanmazoyer: idem

Asterix/utils/gaussians.py
johanmazoyer: ok

Asterix/wfsc/__init__.py
johanmazoyer: ok

Asterix/wfsc/corrector.py
johanmazoyer: here we are in the same folder, we're not in Asterix root. correcged the import order

Asterix/wfsc/corrector.py
johanmazoyer: ok, I correct both

Asterix/wfsc/corrector.py
johanmazoyer: corrected

johanmazoyer:
> I think this looks pretty good, but I do have the feeling like it is a little over-done at this point. I think (?) the main goal is to group already existing but slightly concealed groups better, not necessarily to force everything that exists in Asterix into a group - it is ok to leave thins in the top-level location if there just is not consistent group (yet).
> 
> Below are my suggestions for the reorganization, all at your own discretion. I left a couple of inline comments but I will pay more attention to that in the next round once the new structuring is set.
> 
>     * I recommend using this PR to rename the folder `Model` to `model`
Already done

>     * I very much disagree with the need for the folder `loop`. The name alone isn’t very specific and the module `correction_loop.py` can easily be moved to `wfsc`, while `save_results.py` can be moved to `utils` or even the top-level folder. 
ok, I will correc

>     * I am not sure I agree with `[maskDH.py](http://maskDH.py)` being in the folder `wfsc`. One could need to use this to define a DH without running closed-loop WFS&C. I see that the parameters for the DH are being passed in from `Correctionconfig`, that’s the only tie to wfsc here. Even so, I would probably rather put it in the top-level package since there is simply nothing comparable to this right now.
Not an easy choice, but I prefer to put it here than to leave it in the base folder
 
>     * I also find the separation between the fodlers `optics` and `optical_systems`a little arbitrary. The only two modules that are currently in `optics` have to do with aberrations or with propagations. I would almost say to just resolve this folder and put these two modules back into the top-level package. If there is ever a new module to be made that touch aberrations or propagations, we can always regroup then. And this would also leave you the name.
> 
>     * If you do the above point, you could call `optics` what is now called `optical_systems`. Always better to have single-word folders.
ok... need to think about it

Asterix/tuto_asterix_model.py
ivalaginja: Then I might have misunderstood: I thought we'd implement linting "light", meaning just running to see what's amiss, not actually reformatting code, I agreed with you when we talked about it I would rather only let humans do/trigger that.
What I am saying here is that if it already is nicely formatted (indented in this case), a code change should preserve that. Otherwise you are just going back and forth between enforcing a certain formatting and not enforcing it.

Asterix/loop/correction_loop.py
ivalaginja: I see. Learned something new!

Asterix/tuto_asterix_model.py
johanmazoyer: not sure to understand what you mean

johanmazoyer: i care about indenting if it might send an error (if / for loops, etc). I don't care in the function parameters like here and I'll ran a full yapf formating for all functions in a separate PR after

ivalaginja: But... why do a separate PR for formatting if you could just run it now?

ivalaginja: Remember how `master` is always supposed to have a working version of the code? "working" in this context would also mean in the default format - so as much as broken code should not be merged, mis-formatted code shouldn't be merged. Especially easy if you have an automatic formatter, no? :)

Asterix/wfsc/wf_sensing_functions.py
```diff
@@ -0,0 +1,210 @@
+# pylint: disable=invalid-name
+# pylint: disable=trailing-whitespace
+
+import numpy as np
+
+from utils import invert_svd
```
ivalaginja, line 6: This should be added in the line below and be `from Asterix.utils import resizing, invert_svd`.

source/create_my_testbed.rst
```diff
@@ -155,20 +155,20 @@ and ``todetector_intensity`` documentation in :ref:`os-label`
...
-    config = useful.read_parameter_file(parameter_file)
+    config = read_parameter_file(parameter_file)
     modelconfig = config["modelconfig"]
     DMconfig = config["DMconfig"]
 
-    DM1 = OptSy.DeformableMirror(modelconfig,
+    DM1 = DeformableMirror(modelconfig,
                                     DMconfig,
                                     Name_DM='DM1',
                                     Model_local_dir=Model_local_dir)
```
ivalaginja, line 174: This is not code so your yapf run didn't catch it 😬

source/create_my_testbed.rst
```diff
@@ -195,29 +195,29 @@ It can only be initialized by giving a list of Optical Systems and it will creat
...
     DMconfig = config["DMconfig"]
 
-    pup_round = OptSy.Pupil(modelconfig)
+    pup_round = Pupil(modelconfig)
 
-    DM34act = OptSy.DeformableMirror(modelconfig,
+    DM34act = DeformableMirror(modelconfig,
                                     DMconfig,
                                     Name_DM='DM1',
                                     Model_local_dir=Model_local_dir)
```
ivalaginja, line 211: Like above

source/create_my_testbed.rst
```diff
@@ -195,29 +195,29 @@ It can only be initialized by giving a list of Optical Systems and it will creat
...
 
-    DM32act = OptSy.DeformableMirror(modelconfig,
+    DM32act = DeformableMirror(modelconfig,
                                     DMconfig,
                                     Name_DM='DM3',
                                     Model_local_dir=Model_local_dir)
```
ivalaginja, line 216: Another

source/create_my_testbed.rst
```diff
@@ -195,29 +195,29 @@ It can only be initialized by giving a list of Optical Systems and it will creat
...
 
-    corono = OptSy.Coronagraph(modelconfig, Coronaconfig)
+    corono = Coronagraph(modelconfig, Coronaconfig)
     # and then just concatenate
-    testbed = OptSy.Testbed([pup_round, DM34act, DM32act, corono],
+    testbed = Testbed([pup_round, DM34act, DM32act, corono],
                             ["entrancepupil", "DM1", "DM3", "corono"])
```
ivalaginja, line 221: One more

source/create_my_testbed.rst
```diff
@@ -227,16 +227,16 @@ just like that:
 
 .. code-block:: python
 
-    testbed = OptSy.Testbed([pup_round, DM34act, DM32act, DM32act, corono],
+    testbed = Testbed([pup_round, DM34act, DM32act, DM32act, corono],
                         ["entrancepupil", "DM1", "DM3", "DM4", "corono"])
```
ivalaginja, line 231: And one more

source/create_my_testbed.rst
```diff
@@ -227,16 +227,16 @@ just like that:
...
 
 or a specific pupil in the entrance plane of the coronagraph (e.g. like the Roman configuration).
 
 .. code-block:: python
 
-    pup_roman = OptSy.Pupil(modelconfig, PupType = "RomanPup")
-    testbed = OptSy.Testbed([pup_round, DM34act, DM32act,pup_roman, corono],
+    pup_roman = Pupil(modelconfig, PupType = "RomanPup")
+    testbed = Testbed([pup_round, DM34act, DM32act,pup_roman, corono],
                                 ["entrancepupil", "DM1", "DM3", "romanpupil" , "corono"])
```
ivalaginja, line 240: This might be the last ^^

source/estimation.rst
```diff
@@ -77,10 +81,10 @@ resized to the ``Estim_sampling``:
 
 .. code-block:: python
 
-    import Asterix.processing_functions as proc
+    from Asterix.utils import resizing
     # testbed is previously defined
 
-    resultatestimation = proc.resizing(testbed.todetector(voltage_vector=init_voltage,
+    resultatestimation = resizing(testbed.todetector(voltage_vector=init_voltage,
                                     entrance_EF=input_wavefront),myestim.dimEstim)
```
ivalaginja, line 88: Ah one more

source/index.rst
```diff
@@ -56,131 +56,143 @@ main_THD.py
     :direction: horizontal
 
 
-correction_loop.py
+correction_loop
 -----------------
-.. automodule:: Asterix.correction_loop
+.. automodule:: Asterix.loop.correction_loop
```
ivalaginja, line 61: This is in `correctoin_loop` now.

source/index.rst
```diff
@@ -56,131 +56,143 @@ main_THD.py
...
     :members:
+    :show-inheritance:
+
+.. automodule:: Asterix.loop.save_and_read
+    :members:
+    :show-inheritance:
+
 
-.. callgraph:: Asterix.correction_loop.correction_loop
+.. callgraph:: Asterix.loop.correction_loop.correction_loop
```
ivalaginja, line 70: Idem

ivalaginja (changes requested):
The new structure looks good! I have found one import error and a couple of formatting mishaps in the docs (sorry ^^), but also some uncaught changes in the docs from your most recent commits. Should all be very fast to fix.

Have you run the example script to make sure everything still runs ok with all these changes?

Asterix/optics/phase_amplitude_functions.py
```diff
@@ -1,7 +1,9 @@
+# pylint: disable=invalid-name
+# pylint: disable=trailing-whitespace
+
 import numpy as np
-import Asterix.propagation_functions as prop
-import Asterix.processing_functions as proc
-import Asterix.save_and_read as useful
+from .propagation_functions import mft
```
ivalaginja, line 5: Found one more: you have one relative import left over, this should be an absolute import.

For some resources on relative vs absolute imports (there's many more to be found through googling):
https://stackoverflow.com/a/16748366/10112569

In essence, absolute imports are to be preferred because:
- They future-proof your modules against future rearrangements, needing to change less in that case (and hence breaking less).
- You can use said modules in detached structures like notebooks, which break easily or don't work at all with relative internal imports because they can lead to circular imports.
- It makes it immediately clear where something is coming from.

The only advantage of relative imports that the lines are a little shorter, but Asterix is not a very complex package so it doesn't give you much.

Btw you could make all imports within a package relative imports by just using enough dots (e.g., `from ..utils import xyz`) but please don't make me regret I told you this ^^ This is just to show that the argument of modules living in the same folder is not a strong one in order to use relative imports.

ivalaginja:
Btw it seems like the folder `Model` is still capitalized.

johanmazoyer:
> Btw it seems like the folder `Model` is still capitalized.

Not in mine... Maybe we need to modify it directly in git hub ?

johanmazoyer:
ok, I have applied your modification. I have used that to change the directory name:
https://www.patrick-wied.at/blog/rename-files-and-folders-with-git
there is a trick for case sensitive folders

However, it really seams that doing that there was going to duplicating of the .fits, including the super large wrap vortex .fits file, so I took the decision to commit is out of existence. If you need to use it until your code for replicating it is working, just grab it locally

johanmazoyer:
ok, all your modif have been taken into account and Model -> model 
I have also removed the ----------- bar enclosing the numpy doctrings, hope that makes yo happy :-p
I have made my best to clean the pdf, not an easy task, there is still a lot I don't undersand about sphinx stuff but it's a lot better

johanmazoyer:
the notebook was not working because half the path concatenation was in the form of "path + file" and the other os.join.path(path, file) and therefore depended if you wrote path = 'toto/' or 'toto'. fixed that 

I removed the non nb tutorial, now redundant and requires time to maintain both

+ a few mistakes that appears in testing.
NB tutorial + examples now running smootlhy.

ivalaginja:
> * the notebook was not working because half the path concatenation was in the form of "path + file" and the other os.join.path(path, file) and therefore depended if you wrote path = 'toto/' or 'toto'. fixed that

Yeah I saw issues with this too in other parts of the package. Straight concatenation is a bit faster but not enough I think to chose it over the safer `os.path.join()`. Good catch.

ivalaginja:
@johanmazoyer I agree with all changes but the addition of `ipykernel` to the requirements. It suffices to add the line `%matplitlib inline` to the top of the notebook to make the plots show up:
```python
# Imports
import os
import matplotlib as mpl
%matplotlib inline
from matplotlib.colors import LogNorm
import matplotlib.pyplot as plt
import numpy as np
```

Can you confirm this works for you too, without having to explicitly install `ipykernel` ?

ivalaginja:
Also, the `ipykernel` requirement is already satisfied with the `jupyter` requirement, hence it is not necessary in any case.

johanmazoyer:
Agreed about ipykernel: mistake from my part

johanmazoyer:
- remove another few direct path concatenations
- better printing to understand when we need to wait and why
- yapf formated

Ready for me

ivalaginja (approved):

PR #63, opened 2022-09-24
Clean up `runthd2()`

ivalaginja:
- Between `main_THD.py` and `example_run_asterix.py` , I cleaned out comments and docstrings, unified and simplified variable names.
- I made the function `runthd2()` return the results dictionary. This makes it easier to access the result information outside of the function itself when you want to reuse them programmatically and not read from disk.
- I originally wrote a set of pytests for the main WFS&C loops; they are regression tests and compare their results to results obtained on master and dumped into a text file. But this only works if the interaction matrix is unchanged, which becomes complicated with the current way data directories are handled, so I dropped this for now. I still left in the helper functions for the tests so that they don't get lost.

I am stopping my work here for the sake of small units of work and will do other stuff that doesn't touch this directly on a different branch.


johanmazoyer:
Hi, this is not exactly the direction I would have like Asterix to go to. I always saw the runthd2() as kind of an 'example with the thd2 testbed'. I think that the power of the code is to give you tools to build your test bed in a few line and and define one or several correctors /estimator, easily add new ones and compare their results. In that sense, returning the result at the end of the runthd2() is not how you should do it if you want to use the result of the loop. If you want to continue using the results of the loop, just copy the runthd2() in a new file and modify it as you see fit: there is already a function that return the result of the loop, it's correction_loop(). In that sense the runthd2() is a bit "out of Asterix" imo, it's a collection of Asterix functions you can use if you want to reproduce THD2. 

I shared with you a private repo to show you the kind of thing I currently do to simulate irdis in long pose : see run_cdi_ao_phase_in_parallel.py. I have created a irdis testbed and played with estimation/correction, even though I had to adapt it a little bit to be able to do probes in long exposure time.

On the other side Raphael and Pierre are not very fluent in python and insisted that they need a main with a parameter file (in those exact words) so everything can be used as a black box for THD2, which is why there is this runthd2() existing. There's always been a tension in the process, they insist that I make the code more THD2 oriented (and it's important so that we can more and more precisely reproduce the testbed) and I try to keep the modularity of the code so that me or someone else can easily do something else in the future (SPHERE or else).

So all that to say, this is a very small change and if you make a good argument about returning the result of the code of the runthd2(), I'll of course allow it ! 

Great jobs for the tests, I really like what your trying to do here. WFSC are hard to test because small difference in the entrance phase / way we simulate the matrix will lead to different resutls but can be equally right. Maybe we could have a test with a given hardcoded configuration fqpm / stroke min /etc and ensure that the correction converge to someting below a certain ccontrast value (with this configuration we should normally reach 10-XX if everything is working well). the Param_configspec.ini have default value, maybe we could use that ? We launch a main with an empty parameter file and, every parameter value should go to default.

ivalaginja:
Ok. I have reverted those changes. This is a bit of a useless PR now because it brings neither a functional change nor does it add tests, but it still cleans up a lot of stuff.

That's exactly the kinds of tests I have been thinking about and started setting them up. I would have needed the functional change in `runthd2()` to be able to set it up, otherwise I am duplicating a massive amount of code. But there are other ways.

Yes I build the tests from the couple of standard cases we talked about the other day, and yes we'd need to always be using the same interaction matrix to get repeatable results. For this, we would have to agree on committing one of them to the repo, as an exception to the rule - we should talk about it. But not in here, as it has nothing to do with this PR.

Let me know if there is anything else.

source/run_asterix.rst
```diff
@@ -16,8 +16,8 @@ with the THD2 testbed, just run:
 .. code-block:: python
 
     import os
-    from Asterix import main_THD, Asterix_root
-    main_THD.runthd2(os.path.join(Asterix_root, 'Example_param_file.ini'))
+    from Asterix import main_THD, Asterixroot
+    ain_THD.runthd2(os.path.join(Asterix_root, 'Example_param_file.ini'))
```
johanmazoyer, line 20: forgot 'm' ?

johanmazoyer (approved):

PR #66, opened 2022-09-26
Add save all planes option to main

johanmazoyer:
Added a save_all_planes_to_fits option to mainTHD2 to help debug (Raphael asked for it). I carefully selected what is saved by default if you activate it to avoid sending thousands of .fits per second:
- Mask
- PW Estimate
- 1 full run through testbed (~17 fits for 1 DM testbed -one before and after each planes more or less-).

What we are **not** saved by default, but can be easily done if adding save_all_planes_to_fits=True individually to these functions:
- PW difference (2*nb_probes*17 fits for 1 DM testbed)
- PW Matrix (nb_probes*17 fits for 1 DM testbed)
- EFC Matrix (nb_actuator*17 fits for 1 DM testbed)

PR #67, opened 2022-09-26
Remove circular import risks in Asterix

johanmazoyer:
fixed the circular imports that appear in the `optics` module. 
Solution was to avoid **within the module** broad call to the module, only to part of it
For that reason, within the module `xxx`, we should **not** import with "`from xxx import YYY`" which import all `xxx` package, and prefer `import xxx.yyy as yyy`  and then use `yyy.YYY`, because the second solution does not do a full import of `xxx`

Did the same when necessary to other sub-package at risk (wfsc) to avoid any problems in the future

### DIFF
--- a/Asterix/__init__.py
+++ b/Asterix/__init__.py
@@ -3,6 +3,4 @@ import os
 Asterix_root = os.path.dirname(os.path.realpath(__file__)) + os.path.sep
 model_dir = os.path.join(Asterix_root, "model") + os.path.sep
 
-from .utils.hci_metrics import *
-
 __version__ = "2.4"
--- a/Asterix/example_run_asterix.py
+++ b/Asterix/example_run_asterix.py
@@ -24,7 +24,6 @@ main_THD.runthd2(parameter_file_path,
                      "Nbmode_corr": [320, 340]
                  },
                  NewSIMUconfig={'Name_Experiment': "My_first_experiment"},
-                 save_all_planes_to_fits=False,
                  dir_save_all_planes=None)
 
 print('time correction 1DM perfect estim efc', time.time() - start_time)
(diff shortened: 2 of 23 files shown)


ivalaginja:
This looks like it branched off of #66 so I will only look at it once that is merged.

ivalaginja:
@johanmazoyer I see that you did not change this consistently throughout the package, shouldn't that be done? Otherwise we might get circular import warnings in the future, plus our imports are done inconsistently across the package.

johanmazoyer:
Hum I tried to do it constistenly. 
I remove the broad import in the __setup__ of Asterix package, and then clean the function in optics and wfsc. For utils, I think there was no problem

ivalaginja:
I am just afraid that just because there is no problem now, it doesn't mean there won't be one in the future. There are a lot of direct imports of the optics classes, too many to list them all, but some examples:
https://github.com/johanmazoyer/Asterix/blob/eac5efd0f181416630160bba7376c79ed36283cc/Asterix/optics/coronagraph.py#L9
https://github.com/johanmazoyer/Asterix/blob/eac5efd0f181416630160bba7376c79ed36283cc/Asterix/optics/pupil.py#L10
https://github.com/johanmazoyer/Asterix/blob/eac5efd0f181416630160bba7376c79ed36283cc/Asterix/wfsc/wf_control_functions.py#L16
and many more. I feel like it's safer to change them all. Also, consistency will lead people that might not know why sometimes you import the module and sometimes the class directly.

Also, don't forget the noteboo.

ivalaginja:
Ha I was looking at the wrong branches left and right xD Don't mind my previous comment.

ivalaginja (approved):
Looking ok and fixes the addressed issue.

Note however how this leaves the imports in a fairly inconsistent way across the package:
If I need the class `Pupil`, sometimes it’s imported as `from Asterix.optics import Pupil` now (e.g. in `main_THD.py`) and sometimes as `import Asterix.optics.pupil as pupil` and then call `pupil.Pupil`, like in `coronagraph.py`.
As Johan says, he just removed the subpackage imports within itself.

PR #70, opened 2022-09-27
Make default location for output data not inside repo

ivalaginja:
Currently, the only way to set the saving location of the output files is by changing the appropriate entry (`Data_dir`) in the configfile. This means that the default location for this is inside the repository, which gets messy easily.

This PR makes the default location the user's home directory as long as the configfile entry is set to `Data_dir = '.'`, where it will automatically create the subfolder `asterix_data`. The user can still use the configfile to set the directory they want the data saved in, so everything stays the same in that regard.

As a plus, people can also set an environment variable on their local machine to control the location of the output data of Asterix, but this is not required.

### DIFF
--- a/Asterix/utils/save_and_read.py
+++ b/Asterix/utils/save_and_read.py
@@ -248,3 +248,37 @@ def from_param_to_header(config):
         for scalar in config[str(sect)].scalars:
             header[str(scalar)[:8]] = str(config[str(sect)][str(scalar)])
     return header
+
+
+def get_data_dir(env_var_name="THD2_DATA_PATH", datadir="thd2_data"):
+    """
+    Create a path to local data directory.
+
+    If the environment variable `THD2_DATA_PATH` exists, this is returned as the full data path and the input 'datadir'
+    is ignored. You can set this individually on your OS.
+    If the environment variable does not exist (default for all new users) but the user adapted the ini file, the
+    configfile entry is returned.
+    If the environment variable does not exist and the Data_dir entry in the ini file is '.', the directory 'datadir'
+    is appended to the user's home directory and returned as an absolute path.
+    On MacOS of user 'myuser' for example, this would return:
+        '/Users/myuser/thd2_data'
+
+    Parameters
+    ----------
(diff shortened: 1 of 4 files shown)


Asterix/utils/save_and_read.py
```diff
@@ -249,3 +250,42 @@ def from_param_to_header(config):
         for scalar in config[str(sect)].scalars:
             header[str(scalar)[:8]] = str(config[str(sect)][str(scalar)])
     return header
+
+
+def get_data_dir(env_var_name="THD2_DATA_PATH", config_in=None, datadir="thd2_data"):
+    """
+    Create a path to the local data directory.
+
+    If the environment variable `THD2_DATA_PATH` exists, this is returned as the full data path and the input 'datadir'
```
johanmazoyer, line 259: Why THD2_DATA_PATH ? This is not specific to THD2, maybe put ASTERIX_DATA_PATH

ivalaginja: Ha very correct! I will change this.

Asterix/utils/save_and_read.py
```diff
@@ -249,3 +250,42 @@ def from_param_to_header(config):
...
+    """
+    Create a path to the local data directory.
+
+    If the environment variable `ASTERIX_DATA_PATH` exists, this is returned as the full data path and the input 'datadir'
+        is ignored. You can set this individually on your OS.
+    If the environment variable does not exist (default for all new users) but the user adapted the ini file and
+        accesses the 'Data_dir' entry, the configfile entry is returned and 'datadir' is ignored.
+    If the environment variable does not exist and the Data_dir entry in the ini file is not passed or set to '.',
+        the directory 'datadir' is appended to the user's home directory and returned as an absolute path. On MacOS of
+        user 'myuser' for example, this would return: '/Users/myuser/asterix_data'
```
johanmazoyer, line 264: I'd prefer absolute path in datadir, not relative to home, that way old behavior remain for Charles and Raphael

johanmazoyer (approved):

PR #73, opened 2022-10-01
Sort results directories with timestamp

ivalaginja:
PR for a suggestion:

The current output structure of experiments run by Asterix is as follows:
```bash
Interaction_Matrices
  |-- several .fits files
Labview
  |-- several .fits files
Model_local
  |-- several .fits files
Results
  |-- exp_dir1
  |-- exp_dir2
  |--...
```
where `exp_dir1` and so forth are strings put into the configuration file or passed to the experiment function at runtime. This means that for each new experiment, a new string must be passed manually, otherwise the results will be overwritten.

While this might be fine (although not recommended because you lose the option to compare to previous runs) when debugging, this can be extremely destructive when several different runs are intended to be kept on purpose for good, because it is easy to forget to change the experiment name.

This PR changes this in such a way that each experiment directory under `Results` gets a time stamp appended at the beginning of its name, to avoid overwriting of files. This will prevent any overwriting as different runs are started and looks like this:
```bash
Interaction_Matrices
  |-- several .fits files
Labview
  |-- several .fits files
Model_local
  |-- several .fits files
Results
  |-- 20221001_09-40-27_exp_dir1
  |-- 20221001_09-43-05_exp_dir2
  |--...
```

I believe the main objection might be that people could plaster their local drive with .fits files while running experiments over and over, but I argue that is is much easier and faster to go into your simulation directory regularly to clean out what you don't need while having the safety that you don't lose work you actually want, by accident, rather than to have to redo certain runs just because you forgot to change the experiment string.

In short, this PR creates the name of the experiment directory from the current date and time stamp, which avoids overwriting of result files.

### DIFF
--- a/Asterix/utils/save_and_read.py
+++ b/Asterix/utils/save_and_read.py
@@ -292,6 +292,17 @@ def get_data_dir(env_var_name="ASTERIX_DATA_PATH", config_in=None, datadir="aste
 
 
 def create_experiment_dir(append=''):
+    """
+    Create the name for an experiment directory.
+
+    Create a timestamp including current year, month and day, as well as hour, minute and second. Add the passed
+    suffix 'append' before returning it.
+
+    Parameters
+    ----------
+    append : string
+        Filename suffix to add to timestamp, default is ''.
+    """
     time_stamp = time.time()
     date_time_string = datetime.datetime.fromtimestamp(time_stamp).strftime("%Y%m%d_%H-%M-%S")
 
--- a/Asterix/utils/save_and_read.py
+++ b/Asterix/utils/save_and_read.py
@@ -256,7 +256,7 @@ def get_data_dir(env_var_name="ASTERIX_DATA_PATH", config_in=None, datadir="aste
     """
     Create a path to the local data directory.
 
-    If the environment variable `ASTERIX_DATA_PATH` exists, this is returned as the full data path and the input 'datadir'
+    If the environment variable `env_var_name` exists, this is returned as the full data path and the input 'datadir'
         is ignored. You can set this individually on your OS.
     If the environment variable does not exist (default for all new users) but the user adapted the ini file and
         accesses the 'Data_dir' entry, the configfile entry is returned and 'datadir' is ignored.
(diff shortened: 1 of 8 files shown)


johanmazoyer:
hi, if i'm not mistaken right now, the files themselves are time stamped,eg:
`Results/My_first_experiment/20220930_21h48m14s_DM_Strokes.pdf`
so it is already not overwriting result files. 
I agree having then in time stamped folders it would be more easier to filter the important result from the stupid test, so I generally agree with this change. Please run this through @rgalicher he will probably have an opinion on that ! 

A few changes : 
- Would we keep the time stamp on the files as well ? I would say probably no.
- Should we time stamp to the second / minute ? the question is: can we do several DH correction inside a minute ? I think the answer is yes, especially if we run a for loop with a lot of DH correction to explore a certain type of parameters (for example, if the matrices are already calculated, it can be quite fast). so i tend to think the folder should be time stamped at the second of creation

Asterix/utils/save_and_read.py
```diff
@@ -288,3 +289,14 @@ def get_data_dir(env_var_name="ASTERIX_DATA_PATH", config_in=None, datadir="aste
         if not os.path.isdir(home_path):
             os.mkdir(home_path)
         return home_path
+
+
+def create_experiment_dir(append=''):
+    time_stamp = time.time()
+    date_time_string = datetime.datetime.fromtimestamp(time_stamp).strftime("%Y%m%d_%H-%M-%S")
```
johanmazoyer, line 296: oh I actually see you are already down to the second then alright ignore my comment !

Asterix/utils/save_and_read.py
```diff
@@ -288,3 +289,14 @@ def get_data_dir(env_var_name="ASTERIX_DATA_PATH", config_in=None, datadir="aste
         if not os.path.isdir(home_path):
             os.mkdir(home_path)
         return home_path
+
+
+def create_experiment_dir(append=''):
```
johanmazoyer, line 294: need to add docstring ! 😱

johanmazoyer: Also I missed it during last PR but get_data_dir docstring is not complete (missing parameters and missing 'default value' for conditional argument)

johanmazoyer:
Need also to modify the manual in Correction loop part, in correction.rst

Asterix/utils/save_and_read.py
ivalaginja: Will fix this, thanks for pointing it out!

ivalaginja:
Thanks for looking over it, I completely ignored the fact that the files are now time-stamped as well, which is annoying and unnecessary. I will fix this and ping this when I implement the changes.

ivalaginja:
@johanmazoyer ready for re-review, I also pinged @rgalicher on Slack in case he has something to add.

johanmazoyer (approved):

PR #76, opened 2022-10-06
Add flake8 checks

ivalaginja:
This PR does a couple of different things:
- Add the file `.flake8` that specifies what linting rules Asterix will go by.
- It adds a github workflow running a linting job based on this flake8 file.
- Remove all flags disabling certain pylint flags.
- Add `flake8` to the dependencies.
- Fix up all flake8 complaints according to the new flake8 file.

The whole goal would be to activate automatic flake8 check for every PR that gets opened up against master - @johanmazoyer let's talk about it offline. That way, we would get rid of most of the inconsistent reformatting we keep doing manually.

None of these changes actually change the code, they just give you all the flags that tell you what you need to change to comply with the flake8 file.

**A flake8 check can be run locally:**
- `$ conda activate asterix`
- `$ cd Asterix`
- `$ flake8 . --max-line-length=121 --count --statistics`


ivalaginja:
@johanmazoyer would you mind revieweing this under priority? It touches a lot of code and I don't want it to get into the way of more important changes.

johanmazoyer (approved):

.flake8
```diff
@@ -0,0 +1,42 @@
...
+    E265
+    E271
+    E4
+    E7
+    E9
+    W291
+    W292
+    W293
+    W6
+    #F401
```
ivalaginja, line 26: Just fyi, I want to reactivate this one eventually because it flags unused imports. I commented it out for now because it complains about the imports in one of the `__init__.py` files and I haven't figured out yet how to fix this.

PR #79, opened 2022-10-10
Force the `diam_pup_in_pix` parameter to an even value (for now)

johanmazoyer:
Odd number in the `diam_pup_in_pix` parameter lead to `self.prad` value to be non integer.
I thought this was not a problem (I actually released this constraint a few PRs ago ) if the pupil radius is not exactly a pixel number.

The problem came from MFTs that come back to pupil plane, with parameter `dim_output = int(2 * self.prad)` therefore an odd number. When padded to `dim_overpad_pupil` this lead to an offset (odd dimension array padded into an even dimension array). The solution should be simple (`prad` can be non integer but the pupil plane must always be of even size). For now a quick fix is to force the `diam_pup_in_pix` parameter to an even value.

### DIFF
--- a/Asterix/Example_param_file.ini
+++ b/Asterix/Example_param_file.ini
@@ -32,7 +32,7 @@ onbench=True
     diam_pup_in_m = 8.3e-3
 
     # pupil diameter (in pixels)
-    diam_pup_in_pix = 125
+    diam_pup_in_pix = 100
 
     # overpadding pupil plane factor
     overpadding_pupilplane_factor = 2.0
--- a/Asterix/optics/optical_systems.py
+++ b/Asterix/optics/optical_systems.py
@@ -35,6 +35,10 @@ class OpticalSystem:
 
         """
 
+        if (modelconfig["diam_pup_in_pix"] % 2 == 1) :
+            raise ValueError("Please set diam_pup_in_pix parameter to an even number.")
+        
+
         # pupil in pixel
         self.prad = modelconfig["diam_pup_in_pix"] / 2
 


johanmazoyer:
Ok, I thought it over and I think that there is really no reason to EVER have an odd number of pixel in the pupil. 
1. if we use a round pupil, then we do not really care, it's just a parameter that helps speed basically and 124 or 125 wont change a thing
2. If we use a .fits file we might want to use a odd number if it is a divisor (this is actually the reason I started using it, because 125 was a divisor of 500). But then in that case, we will not be able to padd this odd number pupil in an even array in a satisfactory way (the whole goal of using an integer is to be able to rebin and avoid interpolation).

All that to say that this correction will stay, this is the good way to do that.

Asterix/optics/optical_systems.py
```diff
@@ -35,6 +35,9 @@ def __init__(self, modelconfig):
 
         """
 
+        if (modelconfig["diam_pup_in_pix"] % 2 == 1):
```
ivalaginja, line 38: I don't think you need the parentheses here.

Asterix/Example_param_file.ini
```diff
@@ -32,7 +32,7 @@ onbench=True
     diam_pup_in_m = 8.3e-3
 
     # pupil diameter (in pixels)
-    diam_pup_in_pix = 125
+    diam_pup_in_pix = 100
```
ivalaginja, line 35: Is there a particular reason for this choice? I just put in 126 when I tested it but I guess 100 px is just as good.

johanmazoyer: Yes actually if you are using the Roman pupil it open the 500x500 .fits file and then rebin it. It throws an error if diam_pup_in_pix is not an even divisor (100, 250, 500) (and before this PR, 125 was also a good option). For a clear pupil, this is not important but it's probably better to put 100 at the default option for this reason, you can change it if you want to 126 is you want

ivalaginja (approved):

PR #81, opened 2022-10-11
format docstring with docformatter

johanmazoyer:
Found a automatic formater for docstring which removes trailing whitespace
https://github.com/PyCQA/docformatter

reintroduced the trailing whitespaces error in the flake8. temporaly until we can discuss it. I'm quite confident that Raphael won't have the patience to remove trailing whitespaces...


ivalaginja:
@johanmazoyer what is the purpose of this PR? It does not give new guidance or process on formatting, it just introduces a bunch of changes across the repo.

I really couldn't care less about which convention we follow, the point is simply that there needs to be a reproducible, formal guideline. If this means you want everybody to install this new tool and run it on all of their PRs, then please also add it to the environment file and setup.py, together with an added note in the README on the PR procedure. `docformatter` enforces PEP 257 while flake8 flags PEP8 incompliances, so they don't do the same thing, they just happen to solve a very specific local problem. The reporoducibility is key here.

johanmazoyer:
The goal is for me to find a way to remove trailing whitespaces in docstring automatically (yapf does it in the code, not in the docstring). 

I don't want to do care about trailing whitespaces, it's fastidious (it took me 3 iterations to remove them all in the MFT PR) and cannot ever result in an error in the code and does not even improve readability. So I first removed it completely in the flake8 (commented W291 and W293) and then realized that there was maybe a tool to do it automatically, so i could run it on the entire directory without caring. 

I'm honestly really in favor of removing these errors completely, this is really a deterrent for people for developing for Asterix. I'm 100% sure that Raphaël will laugh when you tell him that he cannot do a PR without removing all trailing whitespaces. More generally, let's go one error after another in the flake8, and decide together which ones to keep (we can sort them in dangerous / unimportant but quick to correct / unimportant and time consuming). I agree to keep a clean code, but we cannot be stricter than all astronomy pipelines or nobody will ever come code with us. It's not our job to educate our colleagues on how to code properly and those who will be annoyed by that will just ignore this package entirely.

PR #82, opened 2022-10-11
reformat all Asterix to 121 line length

johanmazoyer:
I realized that if my autoformater was modifying a lot of your lines it was because it was set to 110 line length and not 121, which is the line you are using. 
I modified it to 121 and run the code. This should happens a lot less now that we agree on line length.

This is 100% a formatting PR, no change of the code was done

### DIFF
--- a/Asterix/optics/deformable_mirror.py
+++ b/Asterix/optics/deformable_mirror.py
@@ -66,8 +66,8 @@ class DeformableMirror(optsy.OpticalSystem):
         else:
             # first thing we do is to open filename_grid_actu to check the number of
             # actuator of this DM. We need the number of act to read and load pushact .fits
-            self.total_act = fits.getdata(
-                os.path.join(model_dir, DMconfig[self.Name_DM + "_filename_grid_actu"])).shape[1]
+            self.total_act = fits.getdata(os.path.join(model_dir,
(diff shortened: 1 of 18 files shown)


Asterix/optics/deformable_mirror.py
```diff
@@ -289,10 +288,12 @@ def creatingpushact(self, DMconfig):
...
-                simu_grid[1, i] = simu_grid[1, i] * np.cos(np.radians(angerror)) - simu_grid[0, i] * np.sin(
-                    np.radians(angerror))
-                simu_grid[0, i] = simu_grid[1, i] * np.sin(np.radians(angerror)) + simu_grid[0, i] * np.cos(
-                    np.radians(angerror))
+                simu_grid[
+                    1,
+                    i] = simu_grid[1, i] * np.cos(np.radians(angerror)) - simu_grid[0, i] * np.sin(np.radians(angerror))
+                simu_grid[
+                    0,
+                    i] = simu_grid[1, i] * np.sin(np.radians(angerror)) + simu_grid[0, i] * np.cos(np.radians(angerror))
```
ivalaginja, line 296: 😭

ivalaginja (approved):
LGTM

PR #85, opened 2022-11-02
Create separate THD2 class

ivalaginja:
The THD2 testbed can be created by concatenating its individual optical elements into a `Testbed` class, which creates a top-level `OpticalSystem`. To run something with `runthd2()`, the only place this is done within Asterix right now is inside of `main_THD.py`.

To be able to create the standard THD2 setup from the configfile elsewhere without copy-pasting code, I wrote a small class called `THD2` that inherits from `Testbed` and simply feeds it what you need to get the THD2 setup.

All example runs have the same results like on master.


ivalaginja:
Thoughts on this @johanmazoyer?

ivalaginja:
Yo @johanmazoyer about this. This PR doesn't actually add anything new, it just wraps the THD simulations into a class to make it easier to call it. Asterix was doing THD simulations before anyway, so this is not that big a change, it just makes it a lot easier to use. If you are unsure about mangling this with the basic optical elements, I could for example rename the new module I made `simulators.py` instead of `thd.py`. Let me know what you think.

Asterix/optics/thd.py
```diff
@@ -0,0 +1,58 @@
...
+
+from Asterix.utils import get_data_dir, read_parameter_file
+from Asterix.optics import Pupil, Coronagraph, DeformableMirror, Testbed
+
+
+class THD2(Testbed):
+    """Testbed object configured for THD2, from input configfile.
+
+    AUTHOR : ILa
+
```
johanmazoyer, line 11: add creation date

johanmazoyer (commented):
Hi, I am convinced that this is a good, idea, thanks. I am a bit concerned because waht you have defined is not really the thd2, but a 1/2 DM coronagraphic testbed, associated at "example" parameter file. But if you change this file, it will not be the thd2 testbed.
I would not mind fixing in this class the thd parameters that will not change. I am thinking 
- filename_instr_apod = "Clear"
- DM properties (size, actuator influence function, DM1_z_position, actuator grid positions, etc)
- Science_sampling
What do you think ?

ivalaginja:
Excellent point! I will do that.

ivalaginja:
I just looked at it again, and while I agree with your comment that this new class only represents THD2 if the parameter file is set up to be the configfile, it is also true that the parameterfile is an input parameter to the init method of the class. Also, it has been like this before anyway when running THD experiments, so this class just follows what is already common practice inside this package.

Refactoring this would take more time because we would have to start ripping things apart in terms of what is pre-set and what is an input (DM configuration vs. coronagraph configuration vs overall configuration, and how all of these get passed around). We can do this step by step, I am not against it, but this PR really aims at just simplifying the use of what is already there, not changing it. If I start using hard-coded things inside this class, I completely take away the advantages of the current setup with the input file and this class could become obsolete very fast. Or we have to keep changing it.

Does this make sense? I would like to separate the concept of "modularize what is already there" (in this PR) from "improve on the definition of individual simulators", which we can tackle separately.

ivalaginja:
@johanmazoyer per Slack on 9 Nov 2022:
> What do you think about this compromise : we put the thd2 class in the Main_THD.py and we create a separate thd2_param.ini file ?

ivalaginja:
I put it in `main_THD.py` as suggested @johanmazoyer. Note how it's not captured by any submodule import anymore, so I had to expose main_THD in the package-level init file.

All ini files but the example is ignored; should I actually commit and push a new THD parameter file or just keep it locally? Because in the end, for you and me we will just be using the main example parameter file anyway then, so maybe there's not even a need for a new file.

ivalaginja:
This is ready for review @johanmazoyer

johanmazoyer (approved):
good for me

PR #87, opened 2022-11-02
Configure CI

ivalaginja:
This PR adds the configuration file to automatically run pytests when a PR is opened (but not when opened as draft), opened for review and each time a review is requested (when the yellow dot besides a reviewer is made to appear).

This will make the tests run, but we still need this check to be set to "required" in the repository settings so that no PR can be merged without them passing @johanmazoyer.


johanmazoyer:
can we do it only when asking for review? it seems redundant to have opening PR + asking for review

ivalaginja:
Yes, but then if I want to rerun the tests I always have to ask for a review. Which is hard when I have already asked you for a review but you haven't given one - in which case I can't re-ask for a review from you. I would have to pick another random person to ask for a review just to run the tests. If it also runs on "open", I can simply convert back to a draft PR and then re-mark as ready and I can run my tests.

Would you like me to deactivate it when asking for a review instead? In practice, it doesn't make much of a difference.

johanmazoyer (approved):

johanmazoyer:
review ok for me, you need to update the branch or rebase to merge.

PR #88, opened 2022-11-02
Use built-in exceptions that are not just `Exception`

ivalaginja:
Replacing all basic exceptions `Exception` with an appropriate built-in exception.

https://docs.python.org/3/library/exceptions.html
https://realpython.com/python-exceptions/

### DIFF
--- a/Asterix/optics/coronagraph.py
+++ b/Asterix/optics/coronagraph.py
@@ -549,7 +549,7 @@ class Coronagraph(optsy.OpticalSystem):
         if self.prop_apod2lyot == "fft":
             maxdimension_array_fpm = np.max(self.dim_fp_fft)
             if len(self.wav_vec) > 1:
-                raise Exception("knife currently not coded in polychromatic fft")
+                raise NotImplementedError("knife currently not coded in polychromatic fft")
         else:
             maxdimension_array_fpm = self.dimScience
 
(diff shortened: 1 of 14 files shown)


ivalaginja:
@johanmazoyer what sort of exception are you trying to catch here?
https://github.com/johanmazoyer/Asterix/blob/010acff313f2168988d6d2c84985651a72575818/Asterix/wfsc/wf_sensing_functions.py#L168

johanmazoyer:
> @johanmazoyer what sort of exception are you trying to catch here?
> 
> https://github.com/johanmazoyer/Asterix/blob/010acff313f2168988d6d2c84985651a72575818/Asterix/wfsc/wf_sensing_functions.py#L168

That date back from Axel's core code. I think he's purpose is to throw a warning when the SVD do not work, not an exception per se. Maybe we should ask him directly what his purpose is to clarify,

ivalaginja:
Catching an exception here makes absolutely sense, I am just trying to refine it to only catch the exceptions we are willing to catch. I can probably test this and figure it out, but if you remember what the exception was that happened here @axelpotier-astro that'd be great.

ivalaginja:
I put this here now: https://github.com/johanmazoyer/Asterix/issues/91
as it is not high priority.

johanmazoyer (approved):

PR #97, opened 2022-11-03
Fix tests

johanmazoyer:
fix and enhance tests. 
The real test for centering is to compare image and transpose, and image and flip, at least to the numerical error level

### DIFF
--- a/Asterix/tests/test_propagations.py
+++ b/Asterix/tests/test_propagations.py
@@ -19,15 +19,10 @@ def test_mft_centering():
     #                                                                                                   == img[4, 4])
     # assert four_equal_pixels, "PSF from MFT is not symmetric in four center pixels."
 
-    np.testing.assert_array_almost_equal(img,
-                                         np.transpose(img),
-                                         decimal=10,
-                                         err_msg="PSF from MFT is not symmetric (transpose PSF != PSF)")
-
-    np.testing.assert_array_almost_equal(img,
-                                         np.flip(img, axis=0),
-                                         decimal=10,
-                                         err_msg="PSF from MFT is not symmetric (flip PSF != PSF)")
+    assert np.allclose(img, np.transpose(img), rtol=0, atol=1e-10,
+                       equal_nan=True), "PSF from MFT is not symmetric (transpose PSF != PSF)"
+    assert np.allclose(img, np.flip(img, axis=0), rtol=0, atol=1e-10,
+                       equal_nan=True), "PSF from MFT is not symmetric (flip PSF != PSF)"
 
 
 def test_butterworth():
--- a/Asterix/tests/test_propagations.py
+++ b/Asterix/tests/test_propagations.py
@@ -14,19 +14,20 @@ def test_mft_centering():
     efield = mft(pup, real_dim_input=pdim, dim_output=pdim, nbres=samp, dtype_complex='complex128')
     img = np.abs(efield)**2
 
-    four_equal_pixels = (img[3, 3] == img[3, 4]) and (img[3, 3] == img[4, 3]) and (img[4, 3]
-                                                                                   == img[4, 3]) and (img[4, 4]
-                                                                                                      == img[4, 4])
-    assert four_equal_pixels, "PSF from MFT is not symmetric in four center pixels."
+    # four_equal_pixels = (img[3, 3] == img[3, 4]) and (img[3, 3] == img[4, 3]) and (img[4, 3]
+    #                                                                                == img[4, 3]) and (img[4, 4]
+    #                                                                                                   == img[4, 4])
+    # assert four_equal_pixels, "PSF from MFT is not symmetric in four center pixels."
 
-    im_and_trans = np.abs(img - np.transpose(img))
-    im_and_flipx = np.abs(img - np.flip(img, axis=0))
-    im_and_flipy = np.abs(img - np.flip(img, axis=1))
+    np.testing.assert_array_almost_equal(img,
+                                         np.transpose(img),
+                                         decimal=10,
+                                         err_msg="PSF from MFT is not symmetric (transpose PSF != PSF)")
 
-    centered = (np.max(im_and_trans / img) < 1e-8) and (np.max(im_and_flipx / img) < 1e-8) and (np.max(
-        im_and_flipy / img) < 1e-8)
-
-    assert centered, "PSF from MFT is not symmetric"
+    np.testing.assert_array_almost_equal(img,
+                                         np.flip(img, axis=0),
+                                         decimal=10,
+                                         err_msg="PSF from MFT is not symmetric (flip PSF != PSF)")
 
 
 def test_butterworth():
--- a/Asterix/tests/test_propagations.py
+++ b/Asterix/tests/test_propagations.py
@@ -14,11 +14,6 @@ def test_mft_centering():
     efield = mft(pup, real_dim_input=pdim, dim_output=pdim, nbres=samp, dtype_complex='complex128')
     img = np.abs(efield)**2
 
-    # four_equal_pixels = (img[3, 3] == img[3, 4]) and (img[3, 3] == img[4, 3]) and (img[4, 3]
-    #                                                                                == img[4, 3]) and (img[4, 4]
-    #                                                                                                   == img[4, 4])
-    # assert four_equal_pixels, "PSF from MFT is not symmetric in four center pixels."
-
     assert np.allclose(img, np.transpose(img), rtol=0, atol=1e-10,
                        equal_nan=True), "PSF from MFT is not symmetric (transpose PSF != PSF)"
     assert np.allclose(img, np.flip(img, axis=0), rtol=0, atol=1e-10,
--- a/Asterix/tests/test_propagations.py
+++ b/Asterix/tests/test_propagations.py
@@ -14,14 +14,19 @@ def test_mft_centering():
     efield = mft(pup, real_dim_input=pdim, dim_output=pdim, nbres=samp)
     img = np.abs(efield)**2
 
-    im_and_trans = np.abs(img - np.transpose(img))
-    im_and_flipx = np.abs(img - np.flip(img, axis=0))
-    im_and_flipy = np.abs(img - np.flip(img, axis=1))
+    four_equal_pixels = (img[3, 3] == img[3, 4]) and (img[3, 3] == img[4, 3]) and (img[4, 3]
+                                                                                   == img[4, 3]) and (img[4, 4]
+                                                                                                      == img[4, 4])
+    assert four_equal_pixels, "PSF from MFT is not symmetric in four center pixels."
 
-    centered = (np.max(im_and_trans / img) < 1e-6) and (np.max(im_and_flipx / img) < 1e-6) and (np.max(
-        im_and_flipy / img) < 1e-6)
+    # im_and_trans = np.abs(img - np.transpose(img))
+    # im_and_flipx = np.abs(img - np.flip(img, axis=0))
+    # im_and_flipy = np.abs(img - np.flip(img, axis=1))
 
-    assert centered, "PSF from MFT is not symmetric"
+    # centered = (np.max(im_and_trans / img) < 1e-6) and (np.max(im_and_flipx / img) < 1e-6) and (np.max(
+    #     im_and_flipy / img) < 1e-6)
+
+    # assert centered, "PSF from MFT is not symmetric"
 
 
 def test_butterworth():
--- a/Asterix/tests/test_propagations.py
+++ b/Asterix/tests/test_propagations.py
@@ -1,6 +1,5 @@
 import numpy as np
 from Asterix.optics import butterworth_circle, fqpm_mask, mft, prop_fpm_regional_sampling, roundpupil
-from Asterix.utils import quickfits
 
 
 def test_mft_centering():
--- a/Asterix/tests/test_propagations.py
+++ b/Asterix/tests/test_propagations.py
@@ -11,7 +11,7 @@ def test_mft_centering():
     pup = roundpupil(pdim, rad, center_pos='b')
 
     samp = 4
-    efield = mft(pup, real_dim_input=pdim, dim_output=pdim, nbres=samp)
+    efield = mft(pup, real_dim_input=pdim, dim_output=pdim, nbres=samp, dtype_complex='complex128')
     img = np.abs(efield)**2
 
     four_equal_pixels = (img[3, 3] == img[3, 4]) and (img[3, 3] == img[4, 3]) and (img[4, 3]
@@ -19,14 +19,14 @@ def test_mft_centering():
                                                                                                       == img[4, 4])
     assert four_equal_pixels, "PSF from MFT is not symmetric in four center pixels."
 
-    # im_and_trans = np.abs(img - np.transpose(img))
-    # im_and_flipx = np.abs(img - np.flip(img, axis=0))
-    # im_and_flipy = np.abs(img - np.flip(img, axis=1))
+    im_and_trans = np.abs(img - np.transpose(img))
+    im_and_flipx = np.abs(img - np.flip(img, axis=0))
+    im_and_flipy = np.abs(img - np.flip(img, axis=1))
 
-    # centered = (np.max(im_and_trans / img) < 1e-6) and (np.max(im_and_flipx / img) < 1e-6) and (np.max(
-    #     im_and_flipy / img) < 1e-6)
+    centered = (np.max(im_and_trans / img) < 1e-8) and (np.max(im_and_flipx / img) < 1e-8) and (np.max(
+        im_and_flipy / img) < 1e-8)
 
-    # assert centered, "PSF from MFT is not symmetric"
+    assert centered, "PSF from MFT is not symmetric"
 
 
 def test_butterworth():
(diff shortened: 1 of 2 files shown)


ivalaginja (changes requested):
Running pytest on this branch gives me three failed tests.
I think it's time to activate the pytests as a branch requirement for PRs in to master in the repo settings @johanmazoyer .

ivalaginja:
Ok now only the MacOS test fails 😆  I will look into it.

Asterix/tests/test_pupil.py
```diff
@@ -1,12 +1,17 @@
...
-    assert zeros_left and center and zeros_right, f"{dim}-pixel pupil is not centered between pixels, in center of array."
+    pupil = roundpupil(dim, rad, grey_pup_bin_factor=2, center_pos='b')
+    centerb = (pupil == np.transpose(pupil)).all() and (pupil == np.flip(pupil, axis=0)).all() and (pupil == np.flip(
+        pupil, axis=1)).all()
+
+    pupil = roundpupil(dim + 1, rad, grey_pup_bin_factor=3, center_pos='p')
+    centerp = (pupil == np.transpose(pupil)).all() and (pupil == np.flip(pupil, axis=0)).all() and (pupil == np.flip(
+        pupil, axis=1)).all()
+
+    assert centerb and centerp, f"{dim}-pixel pupil is not centered between pixels, in center of array."
```
ivalaginja, line 17: We should use full assertions and not just test bools. This was a hack in the direct-pixel-comparison case, but should not be used unless absolutely necessary. See comment in the other test.

Asterix/tests/test_pupil.py
```diff
@@ -1,12 +1,17 @@
...
-    center = (pupil[0, 3] == 1) and (pupil[0, 4] == 1) and (pupil[0, 5] == 1) and (pupil[0, 6] == 1)
-    zeros_right = (pupil[0, 7] == 0) and (pupil[0, 8] == 0) and (pupil[0, 9] == 0)
-    assert zeros_left and center and zeros_right, f"{dim}-pixel pupil is not centered between pixels, in center of array."
+    pupil = roundpupil(dim, rad, grey_pup_bin_factor=2, center_pos='b')
+    centerb = (pupil == np.transpose(pupil)).all() and (pupil == np.flip(pupil, axis=0)).all() and (pupil == np.flip(
+        pupil, axis=1)).all()
+
+    pupil = roundpupil(dim + 1, rad, grey_pup_bin_factor=3, center_pos='p')
+    centerp = (pupil == np.transpose(pupil)).all() and (pupil == np.flip(pupil, axis=0)).all() and (pupil == np.flip(
+        pupil, axis=1)).all()
```
ivalaginja, line 15: These seem fine to me but should be split up into separate assertions.

ivalaginja (changes requested):
I have had some time now to look into the new test in details and I am actually not a fan to replace the old ones with them. I think there is a certain value in keeping the tests simple, even if we're comparing pixels to pixels. In the end, these tests are supposed to test in a way that we would otherwise do manually, and especially when it comes to checking the proper centering of pupils and mfts, we tend to do this on ridiculously small arrays exactly because we want to compare pixels.

How about we just add your new tests as extra tests? But instead of the small 8x8 pixel pupils, we feed them pupils and arrays as we would use in our propagations, for example 512 x 512 pixels? That way we keep the simple tests that are very robust, but we also add the new ones you suggested, but we exploit the fact that we don't need to count pixels anymore and work with more realistic arrays. I think both cases are worth being covered.

I left more detailed comments on how to do this below.

I think this will also solve the failed pytest in this PR because I suspect a varying precision in the residuals of `np.transpose()` is making it break on my machine but not on yours and then let you look over it.

Happy to help if you need any assistance with all these, or if you want me to rewrite it.

johanmazoyer:
Hi, thanks for taking the time to look into this. Please accept PR #98 so that we can have a tool to make the numerical noise of the MFT to discuss this one. 
I disagree that the transponse / flip introduces the numerical noise. These are just pixel swapping function so it is unlikely. Even more, when we do it on simple pupil, we do not see numerical noise introduces. The reason mft - transpose(mft) are not 0 is that MFT are symetrical to the level of the numerical noise. I let you play with this test to convince yourself. 

```
import numpy as np

from Asterix.utils import quickfits
from Asterix.optics.phase_amplitude_functions import *
from Asterix.optics.propagation_functions import *

# normal pupil
pup = roundpupil(60,15, grey_pup_bin_factor=4, center_pos='b')

# test on transpose for this pupil returns true. Unlikely that transpose 
# introduces numerical noise
print("fftpup - transpose eq 0:",(pup - np.transpose(pup) == 0).all())

# measure the max of the PSF on a PSF centered on a pixel
maxPSF = np.max(np.abs(fft_choosecenter(pup, center_pos='bp'))**2)

# measure the FFT PSF
fftpup = np.abs(fft_choosecenter(pup, center_pos='bb')**2)/maxPSF
# fft is not perfectly symetrical, but it is to the numerical error
print("psf_fft - transpose eq 0:",(fftpup - np.transpose(fftpup) == 0).all())
quickfits(fftpup - np.transpose(fftpup), name="psf_fft-transpose")

# measure the MFT PSFs with to level of precision
mftpup64 = np.abs(mft(pup, 30, 60, 30, dtype_complex='complex64')**2)/maxPSF
mftpup128 = np.abs(mft(pup, 30, 60, 30, dtype_complex='complex128')**2)/maxPSF


# MFT are not perfectly symetrical, but numerical noise increase greatly when goign from 128 to 64
print("psf_mft64 - transpose eq 0:",(mftpup64 - np.transpose(mftpup64) == 0).all())
print("psf_fft128 - transpose eq 0:",(mftpup128 - np.transpose(mftpup128) == 0).all())
quickfits(mftpup128 - np.transpose(mftpup128), name="psf_mft128-transpose")
quickfits(mftpup64 - np.transpose(mftpup64), name="psf_mft64-transpose")

quickfits(fftpup - mftpup128, name="psffft - psf_mft128")
quickfits(mftpup64 - mftpup128, name="psf_mft64 - psf_mft128")

quickfits(fftpup, name="psf_fft")
quickfits(mftpup128, name="psf_mft64")
quickfits(mftpup128, name="psf_mft128")
```

johanmazoyer:
This yields a questions 
- beside the test, which is the level of numerical noise we should allow ? complex64 (which leads to float32) should be limiting our results around the 10-10 result I think (more tests necessary). complex64 is a bit faster than complex128 but we do not need this level of speed in our daily simulations. Also labview is in float32 so the matrix multiplication on testbed are limited anyway. TBD next THD meeting.

- More relevant to this test, how to test for PSF symmetry that is relevant on all plateforms. I would argue that the way I am doing it in this test is ok, but adding complex128 option in mft and checking that we are under 10-10. we can use function [numpy.testing.assert_almost_equal](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html)

ivalaginja:
Ok I'll look into this.

Before merging though, could you just change this to not delete the old tests and just have your new ones be added?

johanmazoyer:
> Before merging though, could you just change this to not delete the old tests and just have your new ones be added?

Yes, I can do that, but I'm afraid that if indeed the MFT introduces numerical noise, the pixel to pixel exact comparison (as it was done in the old tests) will be also quite dangerous

Asterix/tests/test_pupil.py
johanmazoyer: ok I'll do that

johanmazoyer: I've separated into 2 separate assertions for the 2 different pupils.
I'm not sure I should split transpose and flip, since they are testing the same thing, the symmetry of the pupil. 
tranpose test that top and left and right and bottom are equal and flip test that left and right are equal. I agree that the last flip is probably an overkill

Asterix/tests/test_propagations.py
```diff
@@ -8,17 +8,26 @@ def test_mft_centering():
     """
     pdim = 8
     rad = pdim / 2
-    pup = roundpupil(pdim, rad)
+    pup = roundpupil(pdim, rad, center_pos='b')
 
     samp = 4
-    efield = mft(pup, real_dim_input=pdim, dim_output=pdim, nbres=samp)
+    efield = mft(pup, real_dim_input=pdim, dim_output=pdim, nbres=samp, dtype_complex='complex128')
```
ivalaginja, line 14: The pixel-comparison test still passes with `complex64`, the test with `np.transpose()` and `np.flip()` doesn't. This supports my idea to keep the pixel-comparison test, we can always add other tests on top, like you did. They simply test for different details.

Asterix/tests/test_pupil.py
```diff
@@ -1,12 +1,17 @@
...
-    assert zeros_left and center and zeros_right, f"{dim}-pixel pupil is not centered between pixels, in center of array."
+    pupil = roundpupil(dim, rad, grey_pup_bin_factor=2, center_pos='b')
+    centerb = (pupil == np.transpose(pupil)).all() and (pupil == np.flip(pupil, axis=0)).all()
+
+    assert centerb, f"{dim}-pixel pupil with center_pos='b' is not centered."
+
+    pupil = roundpupil(dim + 1, rad, grey_pup_bin_factor=3, center_pos='p')
+    centerp = (pupil == np.transpose(pupil)).all() and (pupil == np.flip(pupil, axis=0)).all()
+
+    assert centerp, f"{dim}-pixel pupil with center_pos='p' is not centered."
```
ivalaginja, line 17: Keep it simple. How about this:
```python
def test_roundpupil():
    dim = 10
    rad = 4.3

    pupil = roundpupil(dim, rad, grey_pup_bin_factor=2, center_pos='b')
    assert (pupil == np.transpose(pupil)).all(), f"{dim}-pixel pupil with center_pos='b' is not centered."
    assert (pupil == np.flip(pupil, axis=0)).all(), f"{dim}-pixel pupil with center_pos='b' is not centered."
    assert (pupil == np.flip(pupil, axis=1)).all(), f"{dim}-pixel pupil with center_pos='b' is not centered."

    pupil = roundpupil(dim + 1, rad, grey_pup_bin_factor=3, center_pos='p')
    assert (pupil == np.transpose(pupil)).all(), f"{dim}-pixel pupil with center_pos='p' is not centered."
    assert (pupil == np.flip(pupil, axis=0)).all(), f"{dim}-pixel pupil with center_pos='p' is not centered."
    assert (pupil == np.flip(pupil, axis=1)).all(), f"{dim}-pixel pupil with center_pos='p' is not centered."
```

ivalaginja: When a test fails, we need to know where exactly it fails, hence the separate assertions. And you can't really test too much so I am for testing both flip directions, as you added them.

johanmazoyer: done. I removed the last one because it is redundant : transpose tests that top and left (and right and bottom) are equal and flip tests that left and right are equal. The second flip tests that the top and bottom is already ensured by the first 2

Asterix/tests/test_propagations.py
```diff
@@ -8,16 +8,26 @@ def test_mft_centering():
...
-    four_equal_pixels = (img[3, 3] == img[3, 4]) and (img[3, 3] == img[4, 3]) and (img[4, 3]
-                                                                                   == img[4, 3]) and (img[4, 4]
-                                                                                                      == img[4, 4])
-    assert four_equal_pixels, "PSF from MFT is not symmetric in four center pixels."
+    # four_equal_pixels = (img[3, 3] == img[3, 4]) and (img[3, 3] == img[4, 3]) and (img[4, 3]
+    #                                                                                == img[4, 3]) and (img[4, 4]
+    #                                                                                                   == img[4, 4])
+    # assert four_equal_pixels, "PSF from MFT is not symmetric in four center pixels."
+
+    np.testing.assert_array_almost_equal(img,
```
ivalaginja, line 22: @johanmazoyer is there a reason you wouldn't use
```python
assert np.allclose(img, np.transpose(img), atol=1e-10), "PSF from MFT is not symmetric (transpose PSF != PSF)"
```
from https://numpy.org/doc/stable/reference/generated/numpy.allclose.html ? It's easier to read with the separate `assert`, and since we're only starting with tests might help for clarity.

ivalaginja:
Can we make the pytests a required check again before we merge this, to make sure we don't merge something broken?

Asterix/tests/test_propagations.py
johanmazoyer: No I disagree. You got lucky, it seems that the 4 center pixels seems to be indeed exactly equal in this case. But you have the numerical error appearing when you test for other symetrical pixels : 

`print( (img[2, 3] == img[3, 2]))
print( (img[2, 4] == img[4, 2]))

print(img[2, 3], img[3, 2], img[2, 3] - img[3, 2])
print(img[2, 4], img[4, 2], img[2, 4] - img[4, 2])`

yields: 

```
False
False
415.76685519736236 415.7668551973622 1.7053025658242404e-13
415.76685519736236 415.7668551973622 1.7053025658242404e-13
```
These 2 pairs of pixels that should be exactly equals actually differ at the 13th decimal. 

If we agree that the problem come from the numerical noise of the MFT (and I think this last test proves it, the PSF is not symmetrical at the 13th decimal), then testing on exact equality of a single pixel is dangerous because we might not get lucky next time, those pixels can differ from very small value. The right test is a test using np.testing.assert_array_almost_equal either for pixel value or for transpose. I think that the second one (almost equality of transpose) is better because it test for all pixels : it is a more general test and actually include the first one,

Asterix/tests/test_propagations.py
johanmazoyer: assert_array_almost_equal is a function uses specifically for assert tests, it return a very complete assertion error when it fails : 
Screenshot 2022-11-04 at 11 35 30

But I don't mind if you prefer the allclose one

johanmazoyer:
> Can we make the pytests a required check again before we merge this, to make sure we don't merge something broken?

ok, re-instated the tests !

Asterix/tests/test_propagations.py
ivalaginja: Lol
```python
In [6]: print( (img[2, 3] == img[3, 2]))
   ...: print( (img[2, 4] == img[4, 2]))
   ...: print(img[2, 3], img[3, 2], img[2, 3] - img[3, 2])
   ...: print(img[2, 4], img[4, 2], img[2, 4] - img[4, 2])
```
yields for me:
```bash
True
True
415.76685519736236 415.76685519736236 0.0
415.76685519736236 415.76685519736236 0.0
```

Asterix/tests/test_propagations.py
ivalaginja: I'll leave this decision to you. I just think it might get very confusing to have different test outputs (standard pytest and numpt.testing), especially when we just start rolling out tests. We need to keep things simple.

ivalaginja (changes requested):
@johanmazoyer Suggestion: Let's ditch the pixel-to-pixels tests, but let's use only normal assertions and no np.testing. Then we trigger the tests on this PR and we merge. Ok?

Asterix/tests/test_propagations.py
johanmazoyer: ahah, more reason to avoid testing on the exact equality of pixels. ok let's settle this debate :-D 

```python
for i in range(pdim):
    for j in range(pdim):
        print(i,j,(img[i, j] == img[j, i]), img[i, j], img[j, i], img[i, j] - img[j, i])
```

ivalaginja: Some 1e-6 differences for complex64, some 1e-14 and 1e-15 errors for complex128. I vote for https://github.com/johanmazoyer/Asterix/pull/97#pullrequestreview-1168308255.

Asterix/tests/test_propagations.py
johanmazoyer: Yes that is a good point ,I replace

Asterix/tests/test_propagations.py
```diff
@@ -8,16 +8,21 @@ def test_mft_centering():
...
     img = np.abs(efield)**2
 
-    four_equal_pixels = (img[3, 3] == img[3, 4]) and (img[3, 3] == img[4, 3]) and (img[4, 3]
-                                                                                   == img[4, 3]) and (img[4, 4]
-                                                                                                      == img[4, 4])
-    assert four_equal_pixels, "PSF from MFT is not symmetric in four center pixels."
+    # four_equal_pixels = (img[3, 3] == img[3, 4]) and (img[3, 3] == img[4, 3]) and (img[4, 3]
+    #                                                                                == img[4, 3]) and (img[4, 4]
+    #                                                                                                   == img[4, 4])
+    # assert four_equal_pixels, "PSF from MFT is not symmetric in four center pixels."
```
ivalaginja, line 20: Last request (promise): Can you delete this entirely @johanmazoyer ?

ivalaginja (approved):
Good to go.

Asterix/tests/test_propagations.py
johanmazoyer: I'll open an issue on the complex128/complex64, we need to discuss this

johanmazoyer: but default is now complex128, we need to be safe more than fast right now

johanmazoyer: Do you agree to put atol at 1e-10 then ? I feel this a good contraint but taht allow to variation between computers

ivalaginja: Yes, I would like to increase this a little but for now it's fine. How about you put this question in the new issue too?

ivalaginja:
Tests passed, merging now.

PR #80, opened 2022-10-11
Speeding up MFT !!

johanmazoyer:
Large improvement of the MFT to speed Asterix up

Basically MFT  first measures 2 matrices `AA` and `BB`, and the norm `norm0`. The MFT itself is the matrix multiplication `norm0 * ((AA @ image) @ BB)` (@ is matrix multiplication).

In this PR, I updated MFT so that it can be used in a classical way:
```python
ZeMFT = mft(image,
                real_dim_input=real_dim_input, 
                dim_output=dim_output, 
                nbres=nbres, 
                only_mat_mult=False,
                returnAABB=False) 
```

But also in a new way: You can first measure AA, BB, and norm0 and / or only do the matrix multiplication:
```python
AA,BB,norm0 = mft(image,
                      real_dim_input=real_dim_input, 
                      dim_output=dim_output, 
                      nbres=nbres, 
                      only_mat_mult=False,
                      returnAABB=True) 


ZeMFT = mft(image,
                AA=AA,
                BB=BB,
                norm0 = norm0, 
                only_mat_mult=True,
                returnAABB=False) 
```

By separating those 2 steps you can save a lot of time if you are doing a lot of MFT with the same input and output dimension parameters. Only the second one need to be repeated. This is only done once at the coronagraph initialization and then

 Finally, I realized that previously image was float64/complex64, while AA and BB where complex128. Therefore the results of the MFT was artificially extended to complex128. Now everything is complex64, and it's faster, with almost same results (see next comment). 

Matrix multiplication itself is be done separately in `mat_mult_mft` function. This function can be still enhanced using GPU maybe. 
I tried using [numba](https://numba.pydata.org/) here to save so time but no improvement, word on the internet is that it's not working great for matrix multiplication because our processors are already quite good with that.

PR #98, opened 2022-11-04
introduce choice of complex bit number in MFT

johanmazoyer:
This introduces an MFT parameter to choose the bit number for the complex arrays in the MFT matrices.
Can be complex128 or complex64. The latter increases the speed of the mft but at the cost of lower precision.

I introduced this change because we are currently discussing  the level of numerical noise of the MFT in PR #97 

The change from complex128 to complex64 was made in PR #80 about speeding up the MFT but was maybe not thought enough, so default in now back to complex128

### DIFF
--- a/Asterix/optics/propagation_functions.py
+++ b/Asterix/optics/propagation_functions.py
@@ -138,7 +138,7 @@ def mft(image,
                 terms used in MFT matrix multiplication norm0 * ((AA @ image) @ BB).
     """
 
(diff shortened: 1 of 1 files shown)


johanmazoyer:
Hi, @ivalaginja would you mind accepting this one first, this will help us understand our test problem in the other PR. I have remove the test on pupil until we fixed them

ivalaginja (approved):
Ah I see, this one first. Looks good to me.

PR #99, opened 2022-11-04
Fix warnings from pytest

johanmazoyer:
np.int and np.float are now deprecated https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
This cause warnings in the tests. This PR replace np.int -> int and np.float -> float. No more warning when pytest on my machine


ivalaginja (approved):
LGTM

PR #101, opened 2022-11-04
Fix zoom-in-sampling function for coronagraph propagations

ivalaginja:
Turns out we were cutting off the high spatial frequencies in the outermost propagation plane by increasing the sampling while keeping the array size constant. We know of this now and I will put in some safeguards in later work.


johanmazoyer (approved):
ok for me

ivalaginja:
As info:
This makes the FQPM propagate with a normalized peak intensity of 1.7e-9 and the wrapped vortex of 2.9e-10.

PR #107, opened 2022-11-07
Add tests propagation

johanmazoyer:
Adding 2 propagation tests:
- Compare that MFT and FFT gives the same results, down to the numerical noise
- Test that MFT-1 [MFT [PUP] ] = Pup, down to the numerical noise


ivalaginja (approved):
These are very basic and important tests, it's great to have them.

PR #108, opened 2022-11-08
Add inclination option to wrapped vortex phase mask

ivalaginja:
Depends on #85.

Replaced by #118.

> This PR adds the option to incline the wrapped vortex phase mask along either of the two axes. Below is an example of inclining it by 50 degrees around the x axis (right), compared to no inclination (left):
> Screen Shot 2022-11-08 at 17 21 43

> I also committed the notebooks here that I used to test this, and in which I ran some tests to investigate the effect of a shifted and/or inclined phase mask.

Note how the notebooks used the `THD2` class introduced in #85.

### DIFF
--- a/Asterix/main_THD.py
+++ b/Asterix/main_THD.py
@@ -1,7 +1,7 @@
 import os
 
 from Asterix.utils import create_experiment_dir, get_data_dir, read_parameter_file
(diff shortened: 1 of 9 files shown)


ivalaginja:
Totally fine to converge on #85 first. I'll get going and ping you as I move along.

PR #110, opened 2022-11-11
Simplify photon noise parameters and fix polychromatic photon noise

johanmazoyer:
Fixed several stuff around the photon noise:
- only 1 parameter : nb_photons. If nb_photons=0 no more photon noise
- fix the photon noise in the case of polychromatic correction
- the contrast in the loop is now measured in non-photon noise images because it can lead to  because at some point a lot of values are at 0 and it will artificially lower the contrast. Photon noise is used for the pw images and the photon noise FP intensity is measured and save but not used for contrast.
- Create a normalization to measure photon noise even in non contrast mode.
- create optics.optical_system.add_photon_noise function to measure the photon noise in a FP_image (use to be in todetector_intensity). Making it separate allow us to run the propagation only once in cases where we want both the image with and without photon noise.


Asterix/wfsc/estimator.py
```diff
@@ -244,8 +247,8 @@ def estimate(self, testbed: Testbed, entrance_EF=1., voltage_vector=0., perfect_
             # photon_noise parameter is normally for the whole bandwidth (testbed.Delta_wav). For this
```
ivalaginja, line 247: Comment should be fixed, `photon_noise` does not exist as a parameter anymore.

ivalaginja (changes requested):
Super minor change request, otherwise ready to go.

ivalaginja (approved):

Issue #113, opened 2022-11-16
Reactivate F401 code in flake8 configfile


ivalaginja:
> https://github.com/johanmazoyer/Asterix/blob/fe428422d1627e83b19bf5950d746bc682725df8/.flake8#L26
> Just fyi, I want to reactivate this one eventually because it flags unused imports. I commented it out for now because it complains about the imports in one of the `__init__.py` files and I haven't figured out yet how to fix this.

_Originally posted by @ivalaginja in https://github.com/johanmazoyer/Asterix/pull/76#discussion_r990008453_

johanmazoyer:
Hi, it is solved here
https://stackoverflow.com/questions/31079047/python-pep8-class-in-init-imported-but-not-used

PR #114, opened 2022-11-16
Fix import Error in init.py so "unused import error" can be added in flake8 file

johanmazoyer:
Added a list __all__ of strings with the names of public modules or other objects in the __init__.py for utils
The goal was to solve problem  #113


ivalaginja (approved):
Sweet! Thanks a lot for taking care of this!

PR #116, opened 2022-11-17
Testbed tests

johanmazoyer:
Main goal is to create a testbed_test just to test that creating a simple testbed works
Also in this class 2 small modifs:
- a np.float in deformable_mirror.py that was throwing a deprecation error
- in Main_thd, I renamed parameter_file variable in parameter_file_path to be more precise and not confuse it with the parameter_file itself.

### DIFF
--- a/Asterix/main_THD.py
+++ b/Asterix/main_THD.py
@@ -16,16 +16,17 @@ class THD2(Testbed):
         A read-in .ini parameter file.
     """
 
-    def __init__(self,
-                 parameter_file,
-                 new_model_config={},
-                 new_dm_config={},
-                 new_corona_config={},
-                 ):
+    def __init__(
+        self,
+        parameter_file_path,
+        new_model_config={},
+        new_dm_config={},
+        new_corona_config={},
+    ):
         """
         Parameters
         ----------
-        parameter_file : string
+        parameter_file_path : string
             Absolute path to an .ini parameter file.
         new_model_config : dict, optional
             Can be used to directly change a parameter in the MODELconfig section of the input parameter file.
@@ -36,7 +37,7 @@ class THD2(Testbed):
         """
 
         # Load configuration file
-        self.config = read_parameter_file(parameter_file,
+        self.config = read_parameter_file(parameter_file_path,
                                           NewMODELconfig=new_model_config,
                                           NewDMconfig=new_dm_config,
                                           NewCoronaconfig=new_corona_config)
@@ -59,7 +60,7 @@ class THD2(Testbed):
         super().__init__([entrance_pupil, dm1, dm3, corono], ["entrancepupil", "DM1", "DM3", "corono"])
 
 
-def runthd2(parameter_file,
+def runthd2(parameter_file_path,
             NewMODELconfig={},
             NewDMconfig={},
             NewCoronaconfig={},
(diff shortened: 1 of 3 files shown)


Asterix/main_THD.py
```diff
@@ -16,16 +16,17 @@ class THD2(Testbed):
...
-                 new_dm_config={},
-                 new_corona_config={},
-                 ):
+    def __init__(
+        self,
+        parameter_file_path,
+        new_model_config={},
+        new_dm_config={},
+        new_corona_config={},
+    ):
```
ivalaginja, line 25: Unadapted indentation

Asterix/tests/testbed_test.py
```diff
@@ -0,0 +1,35 @@
+import os
+from Asterix import Asterix_root
+from Asterix.utils import get_data_dir, read_parameter_file
+from Asterix.optics import Pupil, Coronagraph, DeformableMirror
+from Asterix.optics import Testbed as Bench
+# we rename testbed here because the pyhon test thinks it's a
+# test because of the prefix "test" in testbec
```
ivalaginja, line 7: Typo in testbed.

Asterix/tests/testbed_test.py
```diff
@@ -0,0 +1,35 @@
+import os
+from Asterix import Asterix_root
+from Asterix.utils import get_data_dir, read_parameter_file
+from Asterix.optics import Pupil, Coronagraph, DeformableMirror
+from Asterix.optics import Testbed as Bench
+# we rename testbed here because the pyhon test thinks it's a
```
ivalaginja, line 6: Might be worth to capitalize the "testbed" here since that's how the thing you're referring to is written.

Asterix/tests/testbed_test.py
```diff
@@ -0,0 +1,35 @@
...
+                           PupType=model_config["filename_instr_pup"],
+                           angle_rotation=model_config["entrance_pup_rotation"],
+                           Model_local_dir=model_local_dir)
+    dm3 = DeformableMirror(model_config, dm_config, Name_DM="DM3", Model_local_dir=model_local_dir)
+    corono = Coronagraph(model_config, corona_config, Model_local_dir=model_local_dir)
+
+    # Concatenate into the full testbed optical system
+    optical_bench = Bench([entrance_pupil, dm3, corono], ["entrancepupil", "DM3", "corono"])
+
+    optical_bench.todetector_intensity()
```
ivalaginja, line 35: Might be worth to add a basic test for the resulting intensity image. It doesn't have to be anything super smart, but one thing that comes to mind is to test that the resulting arrays of `todetector_intensity()` and the absolute value squared of `EF_through()` are the same.

johanmazoyer: I thought about it. If we want to have any kind of repeatability in the tests here, we need a specifiic `test_parameter_file.ini` file in the `tests` directory and cannot depends on the `Example_param_file.ini` because it might break the test for a lot of reason. For example, and you see it it's perfectly possible to run an image with a coronagraph without pupils and it would run but give very bad results. 

I'll create this `test_parameter_file.ini` in a separate PR when I'll create wfsc tests

Asterix/tests/testbed_test.py
johanmazoyer: done

Asterix/tests/testbed_test.py
johanmazoyer: done

Asterix/main_THD.py
johanmazoyer: done

Asterix/tests/testbed_test.py
ivalaginja: Excellent

ivalaginja (approved):

PR #121, opened 2022-11-22
Make FQPM use zoom-in propagation

### LINKED ISSUE #89: Switch FQPM to use zoom-in sampling propagation as default
This functionality is now available after merging #71, but there are still some issues with it documented in #84, so switching the FQPM over to this too was put on hold.

ivalaginja:
Fixes #89.

This PR changes the FQPM propagations to use the zoom-in sampling. The peak attenuation attained like this is somewhat better than 1e-6 for a 80 pixel round, grey pupil and somewhat better than 1e-7 for a 200 pixel round, grey pupil. I am unsure whether this is sufficient to motivate this change. By comparison the wrapped vortex doesn't do all that much better, the difference is approximately a factor of two.

I also set the pupil to a grey pupil by default in the parameter file. This is needed to make the zoom-in sampled propagation work. There was a question whether this would pose problems for the RST pupils but no, as they gery-scaling is only used for round pupils, not for read-in pupil files.

I also adapted some of the tests. If we don't want to keep the FQPM change after all, I would change this PR so that we can at least merge the updated tests.

If we do change the FQPM, the question arises as whether we should change the Vortex to use the zoom-in propagation too, as it would be the last remaining "perfect" coronagraph by default in Asterix. It yields about the same peak attenuation like the wrapped vortex.

It might be useful to tie the decisions (whether to change to the new propagation function) to a hardware test on THD2.

Needs discussion @johanmazoyer.

### DIFF
--- a/Asterix/Example_param_file.ini
+++ b/Asterix/Example_param_file.ini
@@ -55,7 +55,7 @@ onbench=True
     # if True, all the pupils will be created 10x larger than diam_pup_in_pix and 
     # then rebinned to diam_pup_in_pix at the exception of pupils that are read in 
     # a fit file, which are always rebinned from the sixe of the .fits
-    grey_pupils = False
+    grey_pupils = True
 
 [DMconfig]
 
--- a/Asterix/optics/coronagraph.py
+++ b/Asterix/optics/coronagraph.py
@@ -60,7 +60,7 @@ class Coronagraph(optsy.OpticalSystem):
             # We do not need to be exact, the mft in science_focal_plane will be
 
         if self.corona_type == "fqpm":
-            self.prop_apod2lyot = 'mft'
+            self.prop_apod2lyot = 'regional-sampling'
             self.err_fqpm = coroconfig["err_fqpm"]
             self.achrom_fqpm = coroconfig["achrom_fqpm"]
             self.FPmsk = self.FQPM()
@@ -69,7 +69,7 @@ class Coronagraph(optsy.OpticalSystem):
             else:
                 str_achrom = "nonachrom"
             self.string_os += '_' + str_achrom
-            self.perfect_coro = True
+            self.perfect_coro = False
 
         elif self.corona_type in ("classiclyot", "hlc"):
             self.prop_apod2lyot = 'mft-babinet'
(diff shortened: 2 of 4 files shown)


Asterix/Example_param_file.ini
```diff
@@ -55,7 +55,7 @@ onbench=True
     # if True, all the pupils will be created 10x larger than diam_pup_in_pix and 
     # then rebinned to diam_pup_in_pix at the exception of pupils that are read in 
     # a fit file, which are always rebinned from the sixe of the .fits
-    grey_pupils = False
+    grey_pupils = True
```
johanmazoyer, line 58: I'd prefer to keep the default at False until we have a discussion with Raphael about it

ivalaginja: Yes, I opened this PR as a follow-up to exactly this discussion with @rgalicher and Pierre at the meeting yesterday (I should have mentioned this, sorry). I will find him offline to see what he thinks about this.
Overall both of them seemed ok with it as long as we can confirm it (a) improves the simulations and (b) doesn't make hardware ops worse. This is also why I suggested to maybe change the FQPM to the zoom-in sampling prop (which needs the grey pupil) only after creating a matrix with this that runs fine on hardware.

So if we choose to not go ahead with the change in the grey pupil default, I would also prefer not to change the default propagation for the FQPM, since that would return really bad results; and the FQPM remains the reference coronagraph for THD2.

rgalicher: @johanmazoyer @ivalaginja Ok to upgrade the FQPM to grey pupil + zoom-in sampling prop once it has been demonstrated that the estimation+correction loop works fine (i.e. with same performance as the current version of FQPM). 
Estimation = all estimator (perfect, PW, is there another?)
correction = all correctors (EFC, etc)

johanmazoyer: ok

johanmazoyer (dismissed):

ivalaginja:
I just want to make sure we don't merge this before the hardware tests. The rewritten pytests can wait, so I won't split this PR after all.

PR #125, opened 2022-11-28
Remove type hints from code

ivalaginja:
Something changed in the package or CI setup so that the classes imported for type hinting are now flagged as unused imports, an issue we have not had before. Discovered while working on #124, one failed run here:
https://github.com/johanmazoyer/Asterix/actions/runs/3567127502/jobs/5994434054

So just removing them for now until we either figure something out or just carry on like this.


ivalaginja:
@johanmazoyer it's still failing, you said you still had stuff to push?

ivalaginja:
@johanmazoyer it's still failing, you said you still had stuff to push?

PR #126, opened 2022-11-28
Replace type hints for optical systems to avoid unused import in flake8

johanmazoyer:
Something changed in the package or CI setup so that the classes imported for type hinting are now flagged as unused imports, an issue we have not had before. Discovered while working on #124, one failed run here: https://github.com/johanmazoyer/Asterix/actions/runs/3567127502/jobs/5994434054

Changeing the type hints to something more explicit

### DIFF
--- a/Asterix/optics/testbed.py
+++ b/Asterix/optics/testbed.py
@@ -167,7 +167,7 @@ class Testbed(optsy.OpticalSystem):
 
         for i, DM_name in enumerate(self.name_of_DMs):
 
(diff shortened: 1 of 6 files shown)


ivalaginja (approved):
Nice solution

johanmazoyer:
Please note that type hints requires the use of python >= 3.6 which is not really a problem but should be kept in mind.
More generally, it can be used to flag wrong type error more easily

PR #127, opened 2022-11-29
Full DH fixed

johanmazoyer:
PR #124 actually broke the full DH. This PR fixes that

Before #124 :
options for DH_side: 'top', 'bottom', 'left', 'right' and all others were default to 'full'

Now: 
options for DH_side: 'top', 'bottom', 'left', 'right, 'full' and all others raise an exception


ivalaginja (approved):
Ha good catch. We really need those tests for the WFS&C loops...

PR #128, opened 2022-11-29
Modify thd2 class

johanmazoyer:
If the parameter file is read at several locations (currently in and out of the thd2 class), meant that parameter modification to the config done outside of the THD2 class are not taken into account by THD2 whch re-read the parameter_file and the testbed might even have wrong parameters. It's better to have the parameter only read once and then we only pass the config dictionary object.


Asterix/main_THD.py
```diff
@@ -16,30 +16,19 @@ class THD2(Testbed):
...
-        parameter_file_path : string
-            Absolute path to an .ini parameter file.
-        new_model_config : dict, optional
-            Can be used to directly change a parameter in the MODELconfig section of the input parameter file.
-        new_dm_config : dict, optional
-            Can be used to directly change a parameter in the DMconfig section of the input parameter file.
-        new_corona_config : dict, optional
-            Can be used to directly change a parameter in the Coronaconfig section of the input parameter file.
+        config: dict
+            Parameter dictionary. Must at least contains parameter for Entrance Pupil, Coronagraphs and DM(s).
```
ivalaginja, line 24: Missing white space before colon; I would replace docstring with (fixing typos): "Testbed parameter dictionary. Must at least contain parameters for entrance pupil, coronagraph and DM(s)."

Asterix/main_THD.py
```diff
@@ -16,30 +16,19 @@ class THD2(Testbed):
...
+        Model_local_dir: string
+                path directory to save things you can measure yourself and can save to save time
```
ivalaginja, line 26: Missing white space before colon; and this docstring as it is is pretty confusing unless you already know how the simulator works, so I would suggest: "Directory output path for model-related files created on the file for later reuse.", or something the like.

ivalaginja (changes requested):
Looks good to me, just some minor docstring suggestions.

What needs to be changed before merging is that `notebooks/coronagraph_propagation.ipynb` contains all outputs currently.

Also, in that same notebook it would be much easier to follow what's happening if it actually uses the `THD2` class, but that's not the goal of this PR so feel free to ignore and I ca do that later.

Asterix/main_THD.py
johanmazoyer: About the semi collumn, apparently google style formatting is without semi collumn
https://google.github.io/styleguide/pyguide.html#doc-function-args

Asterix/main_THD.py
johanmazoyer: done

Asterix/main_THD.py
johanmazoyer: updated the comments

ivalaginja (approved):

PR #129, opened 2022-11-29
Wfsc tests

johanmazoyer:
Create a few new tests:
- a param_file_tests.ini which can be used only to perform tests, separate from the defualts parameter file Example_param_file.ini in the Asterix directory. By default this is smaller parameter to go faster. We can optimize the smallest parameter (in pupil plane and focal plane size) to go as fast as possible while having decent performance in contrast. Pupil is 100 pixels by default.
- added polychromatic coronagraphs tests
- for coronagraph tests, remove `center on pixel = True`: almost all coronagraphs propagation in Asterix are centered in between pixel. 
- readjust estimated levels for coronagraph attenuation (now for 100pix aperture and in between 4 pixels) and symmetry levels to work with every OS levels.
- added 1DM (pw / efc) and 2 DM (perfect estim / strokemin) monochromatic correction 

test takes now ~1 minute to run


ivalaginja:
@johanmazoyer I had written some coronagraph tests in #121 already, finally I did decide to split those out so that it does not depend on the decision on the FQPM propagation.

Would you mind having a look at that, it's PR #130. I did it that way so that we can actually test *all* of the asterix coronagraphs without having to duplicate code too much, but this small change would significantly increase the CI test coverage and guard all coronagraphs at least against the most destructive accidental changes. We could then build on top of that,  including your changes in this PR here.

Let me know what you think.

Asterix/tests/test_coronagraph.py
```diff
@@ -37,6 +32,19 @@ def test_all_coronagraphs():
...
+    modelconfig.update({"Delta_wav": 30e-9})
+    modelconfig.update({"nb_wav": 3})
+    for i, coro in enumerate(coros_to_test):
+        
+        # Create the coronagraph
+        corono = Coronagraph(modelconfig, Coronaconfig)
+        try:
+            coro_psf = corono.todetector_intensity(center_on_pixel=True, in_contrast=True)
+        except:
+            assert False, f"Coronagraph '{coro}' does not work in polychromatic light."
```
ivalaginja, line 46: Why did you opt for a try-except here? If the propagation fails, it will throw an error and the test fails anyway.

Asterix/tests/test_wfsc.py
```diff
@@ -0,0 +1,118 @@
...
+import pytest
+import numpy as np
+
+from Asterix.main_THD import THD2
+from Asterix import Asterix_root
+from Asterix.utils import get_data_dir, read_parameter_file
+from Asterix.wfsc import Estimator, Corrector, MaskDH, correction_loop
+
+
+def quick_run_no_save(config, data_dir):
```
ivalaginja, line 12: You might not need this. I think you can just define a test function where you pass `tmpdir`, a pytest ficture that creates a throw-away path for output files for you, and then you can use that as the output path in your tests. The function name would look something like this:
```python
def test_a_wfsc_run(tmpdir):
```
and then you use `tmpdir` as your path.

I haven't tested this for Asterix though, there could be some hickups, but if it works it would be much easier than defining a whole new function.

Asterix/tests/test_coronagraph.py
johanmazoyer: ok, I thought we needed absolutely to have a assert in the test

Asterix/tests/test_wfsc.py
johanmazoyer: Sorry I don't understand. 
I create a new functions so that I can separatly call it for 1 or 2 dm correction

Asterix/tests/test_testbed.py
```diff
@@ -9,18 +10,23 @@
...
-    parameter_file_path = os.path.join(Asterix_root, 'Example_param_file.ini')
+    parameter_file_test = os.path.join(Asterix_root, 'tests', "param_file_tests.ini")
+    test_dir = get_data_dir(datadir="asterix_test_dir")
 
-    model_local_dir = os.path.join(get_data_dir(), "Model_local")
+    model_local_dir = os.path.join(test_dir, "Model_local")
 
     # Load configuration file
-    config = read_parameter_file(parameter_file_path)
+    config = read_parameter_file(parameter_file_test)
```
ivalaginja, line 19: I think is one of the instances where you could indeed do:
```python
def test_def_thd(tmpdir):
    ...
    model_local_dir = os.path.join(tmpdir)
```
and just remove the `test_dir = get_data_dir(datadir="asterix_test_dir")`.

ivalaginja (commented):
Some minor comments, mostly about things we talked about offline.

ivalaginja (approved):

PR #132, opened 2022-11-30
Write out FPM in zoom-in propagation when requested

ivalaginja:
I hadn't implemented any writing out of intermediary planes in the focal plane when we merged the zoom-in sampling, this PR adds the option to write out the FPM when the appropriate keyword is set.

I tested this with the wrapped vortex coronagraph and it works fine.

### DIFF
--- a/Asterix/optics/coronagraph.py
+++ b/Asterix/optics/coronagraph.py
@@ -381,6 +381,17 @@ class Coronagraph(optsy.OpticalSystem):
                 fpm_array = np.ones((self.dimScience, self.dimScience))
             else:
                 fpm_array = FPmsk
+
+                if dir_save_all_planes is not None:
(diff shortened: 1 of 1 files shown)


johanmazoyer (commented):
Hum ok this is not super useful for debugging because the FPmask is easy to get. 

A useful would be to take it inside the prop_fpm_regional_sampling function to have access to FP right before the FPM, which is always interesting to study

ivalaginja:
I agree with you, just wasn't sure how to implement that fast since it needs some extra parameters being passed around.
Btw while working on this I realized this doesn't write out the phase mask, but the phasor, which is not super useful if you want to look at it outside of code. This is consistent with how it's done for the FQPM but I don't think it's super useful. So I am ok with just closing this PR without merging.

PR #133, opened 2022-12-05
error in multichrom correction with perfect estimate

johanmazoyer:
Fix a small error introduced for perfect estimation when I disantangle testbed wl and estimation wl.


ivalaginja (approved):
I am assuming you tested the various cases.

PR #134, opened 2022-12-05
Add notebooks for orientation checks and model inspection

ivalaginja:
This PR just adds some notebooks I thought would be useful on the repo. And fixes a single docstring.

@johanmazoyer I don't think there's a need to review all the notebooks in details but it's up to you.


johanmazoyer (approved):
ok, nb are hard to review online, but do not impact the code behavior anyway

PR #135, opened 2022-12-07
Small error in docs

johanmazoyer:
A mix of a few things:
- a small modifs of a function docstring that was wrong
- a minor improvement in the matrix measurement. I tried to change completely the matrix (so that it could be parallelized) but I did not succeed so it is now commented if I think about something else.
-


ivalaginja (commented):
@johanmazoyer I will approve under the condition you create a new Issue that points to this, so that it doesn't fall under the bus completely.

johanmazoyer:
> @johanmazoyer I will approve under the condition you create a new Issue that points to this, so that it doesn't fall under the bus completely.

done

ivalaginja (approved):

PR #136, opened 2022-12-15
Create read-the-docs website

johanmazoyer:
This push modify the doc using sphinx to have a working readthedoc website. 
Currently accessible via this link 
https://asterix-hci.readthedocs.io/en/update-fro-read-the-docs/
because this is only working in this branch, this will change when it is pushed in master, we will have a more perene address.

- Creation of docs folder
- change a lot of .rst files
- large modification of sphinx configuration conf.py
- creation of readthedocs file
- change in setup.py -> we only install necessary packages and test / doc package can be installed separately

There might be a problem with the pyan graph. I will remove them later if it does not work


docs/todo.rst
```diff
@@ -1,6 +1,9 @@
 .. _todo-label:
 
-**To Do**: These are future improvements to Asterix currenly envisionned (not particularly by order of priority):
+Wanna help ?
```
ivalaginja, line 3: How about you convert all of these into GitHub issues (where possible)? In this way all open points and potential improvements are listed in only one place, and they get updated naturally when they have been worked on.

docs/index.rst
```diff
@@ -0,0 +1,51 @@
...
+Asterix is publicaly available on `GitHub <https://github.com/johanmazoyer/Asterix>`_ and all contributions are welcome!
+
+The development of Asterix is led by Johan Mazoyer with major contributions Axel Potier, Iva Laginja, 
+and Raphaël Galicher from LESIA (Paris Observatory)
+
+.. toctree::
+    :maxdepth: 1
+    :caption: Getting Started
+
+    install
```
ivalaginja, line 17: I don't know where this shows up in the docs, but to save yourself some headaches in the future, I would delete the section "Dependencies". If you really want that info in the docs, I would link to the `environment.yml` or `setup.py` file. Otherwise you just have yet another place where you need to update the dependencies, which is easily forgotten.

ivalaginja:
@johanmazoyer I think you only need 7 dashes under each "Returns", not 8.
https://numpydoc.readthedocs.io/en/latest/format.html

docs/todo.rst
ivalaginja: Also, it might be more helpful to call this "Open issues" or "Backlog" or something. Sounds a little more serious ^^

setup.py
```diff
@@ -30,5 +30,15 @@
     ],
     keywords='Exoplanets imaging high-contrast coronagraphy',
     install_requires=[
-        'astropy', 'configobj', 'flake8', 'jupyter', 'matplotlib', 'numpy', 'pytest', 'scikit-image', 'scipy'
-    ])
+        "numpy",
```
ivalaginja, line 33: Did you un-sort these on purpose?

ivalaginja (commented):
Looks good to me, very cool that you're moving this to read the docs now. I jsut have some minor comments, see below.

Also, it would be nice if the beginner notebook could be displayed in this documentation, if this is possible at all.

docs/index.rst
johanmazoyer: ok, right, I'll do it now

docs/todo.rst
johanmazoyer: ahah ok, I'll do it

setup.py
johanmazoyer: no, I'll reorder

johanmazoyer:
Rebase was a bit trick because files are deleted / moved in the directory, but tests/pylint/sphinx compile seems to work ok ! ok to PR when you are !

johanmazoyer:
> @johanmazoyer I think you only need 7 dashes under each "Returns", not 8. See: https://numpydoc.readthedocs.io/en/latest/format.html#returns

It's a mess. I think it's a minimum of 7, a lot of "returns" where not detected as such initially, but now everything seems to work ok

johanmazoyer:
> Looks good to me, very cool that you're moving this to read the docs now. I just have some minor comments, see below.
> 
> Also, it would be nice if the beginner notebook could be displayed in this documentation, if this is possible at all.

There's a lot of modification possible with the docs now that it's online :
- pyan graphs are not working anymore I think
- Displaying notebooks more easily
- a well displayed documentation also shows very clearly small typos / rst mistakes (bullet lists for example) that need to be corrected 

Let's make this working version public and have other PRs to do more functionnalities

setup.py
```diff
@@ -30,5 +30,15 @@
     ],
     keywords='Exoplanets imaging high-contrast coronagraphy',
     install_requires=[
-        'astropy', 'configobj', 'flake8', 'jupyter', 'matplotlib', 'numpy', 'pytest', 'scikit-image', 'scipy'
-    ])
+        "astropy",
+        "matplotlib",
+        "configobj",
+        "IPython",
```
ivalaginja, line 36: You did indeed reorder, but not alphabetically? 🤔 Also, I think `ipython` should have no caps?

docs/todo.rst
```diff
@@ -1,6 +1,9 @@
 .. _todo-label:
 
-**To Do**: These are future improvements to Asterix currenly envisionned (not particularly by order of priority):
+Open issues
+---------------
+
+These are future improvements to Asterix currenly envisionned (not particularly by order of priority):
```
ivalaginja, line 6: Not going to put any into GitHub issues instead? Just checking.

ivalaginja (commented):
@johanmazoyer can you give me a link to the updated rendered documentation so I can check? The old link gives me the initial version.

setup.py
johanmazoyer: ok, it was an attempt to do it alphabetically :-D

Also, you seem to be right about the caps. Why did the pytest installation worked ?

docs/todo.rst
johanmazoyer: not right no, we need to clean it first / add current best issues

setup.py
ivalaginja: Ah xD I have no clue. Maybe it's makes it all lower caps automatically. Still, worth it to keep things clean here.

ivalaginja:
> @johanmazoyer can you give me a link to the updated rendered documentation so I can check? The old link gives me the initial version.

@johanmazoyer ?

johanmazoyer:
> > @johanmazoyer can you give me a link to the updated rendered documentation so I can check? The old link gives me the initial version.
> 
> @johanmazoyer ?

It's the same link : https://asterix-hci.readthedocs.io/en/update-fro-read-the-docs/
However, you should read it in "private mode" to avoid that your browser read the cache and not the website i think

ivalaginja (approved):

PR #137, opened 2022-12-15
Time-stamped sub folders and commit hash for Labview matrices

ivalaginja:
I would like to save the matrix files created for the Labview controls to individual sub-folders each time they are generated, with each sub-folder getting a time stamp much like the data result folders. This prevents over-writing when running several simulated experiments sequentially. I also added the coronagraph type to the directory name to have at least some initial info, but I am keeping the file names purposefully simple, leaving it up to the user to rename them as needed, since this will likely always be needed.

The second thing added to the sub-folder name is the short commit hash of the repo version the data has been created with.

This will allow us to create a bunch of hardware-ready matrix directories in one go, and trace which commit they have been created on.

Note how this does not distinguish between clean and dirty commits. This means if you are on a commit but you have uncommitted changes, those changes will be reflected in the created matrix but not in the code references (as it's uncommitted changes). We could add a flag for dirty commits but the version here should be sufficient for now.

### DIFF
--- a/Asterix/wfsc/thd_quick_invert.py
+++ b/Asterix/wfsc/thd_quick_invert.py
@@ -23,7 +23,7 @@ def THD_quick_invert(Nbmodes, name_active_DM, matrix_directory, regularization, 
             - Base_Matrix_DM1.fits
             - Direct_Matrix_2DM.fits
 
-    Save the following fits files (that can be read by the testbed controls):
+    Save the following fits files (that can be read by the testbed RTC):
         if DM1 only is active:
             - Matrix_control_EFC_DM1.fits
         if DM3 only is active:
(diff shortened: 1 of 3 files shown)


Asterix/wfsc/thd_quick_invert.py
```diff
@@ -24,7 +23,7 @@ def THD_quick_invert(Nbmodes, name_active_DM, matrix_directory, regularization,
             - Base_Matrix_DM1.fits
             - Direct_Matrix_2DM.fits
 
-    Save the following fits (that can be read by the testbed):
+    Save the following fits files (that can be read by the testbed controls):
```
johanmazoyer, line 26: the testbed RTC

ivalaginja: Fixed

ivalaginja:
> Maybe also copy the commit / date in the header of all fits saved for the computer ?

I was thinking about that, but realistically, we only ever move those files together. The solution with the folder name was faster to implement and we're much likely to look for that than to go in and open the file and get access to the header. It requires all machines who want to see that to have DS9 installed (or something that lets you check the header quickly).

If at some point we do a general restructuring of these files, I'd be happy to add this.

johanmazoyer (approved):

PR #139, opened 2022-12-16
Default image origin to lower left across entire repo.

ivalaginja:
As soon as `Asterix` is imported now, the default image origin for matplotlib will be set to the lower left.

This enforces a uniform image origin across the repo without us having to remember each time when we create new plots. And it will make it more straight forward to compare between images.


ivalaginja:
@johanmazoyer got any time to review this?

johanmazoyer (approved):

PR #140, opened 2022-12-19
Update with new website

johanmazoyer:
A few update on the website:
- better links to cross reference the packages, including a link to the notebook package
- clearer installation for setup and install
- Made reference to the notbebook in the docs. Direct link to notebooks in readthedoc are apparently possible, but I could not make it work myself : https://docs.readthedocs.io/en/stable/guides/jupyter.html
- got rid of the [pyan](https://github.com/Technologicat/pyan) graphs, the package was not maintained anymore and not working directly in readthedocs
- clean the readme and refer direclty to the website to avoid repetition

Comments are welcome to make it clearer and easier to install !


ivalaginja:
Do you want to use this PR to also consolidate the README with the readthedocs website?  Minimum would be to add the link to it in the README, ideally you would get rid of duplicate information and only keep it in the readthedocs.

johanmazoyer:
> Do you want to use this PR to also consolidate the README with the readthedocs website? Minimum would be to add the link to it in the README, ideally you would get rid of duplicate information and only keep it in the readthedocs.

I think it does. The readme have been cleaned, most information have been sent in the install section of the website and we only have the link in the readme. Here is the version that will be online when this PR is pushed : https://asterix-hci.readthedocs.io/en/update_with_new_website/

docs/todo.rst
```diff
@@ -1,8 +1,47 @@
...
+If you want to develop for Asterix, you might need packages to test and build the documentation. Run:
+
+.. code-block:: bash
+    
+    $  pip install -e '.[dev,docs]'
+
+To contribute to Asterix, please follow the following steps:
+    1. Make sure your local `master` branch is up-to-date by pulling.
+    2. Create a new branch off `master` with a name of your choice and commit your work.
+    3. When done, open a PR and request a review after ensuring your branch is up-to-date with the base branch you're merging into (usually `master`) and after running the pytests and the linter locally. You can then the run the pytests
```
ivalaginja, line 15: "You can then the run the pytests" has a "the" too many and I would put a colon after "pytests", like you did after the flake8 linter further below.

Actually, even so it's a little confusing with the previous sentence ("[...] after running the pytest and the linter locally." So maybe Just do:
"[...] after running the pytests and the linter locally as follows. To run the pytests:
```bash
$ pytest
```
To run the flake8 linter:
```bash
$ flake8 . --max-line-length=121 --count --statistics
```
"
It might also be useful to keep the instructions where in the repo these need to be run - I know I like to come back to that because I need to refresh my memory sometimes.

docs/todo.rst
```diff
@@ -1,8 +1,47 @@
...
+    .. code-block:: bash
+
+        $ flake8 . --max-line-length=121 --count --statistics
+
+
+    4. Iterate on the review, once it's approved it will be immediately merged.
+
+**Documentation**
+
+Please update the documentation after modifying Asterix. Docstrings must follow the `numpydoc guidelines <https://numpydoc.readthedocs.io/en/latest/format.html>`_.
```
ivalaginja, line 32: I would be stricter, so instead of saying "Please update the documentation *after* modifying Asterix", I would say: "Please update any affected documentation when working on a PR." That way you won't merge a PR without also having the docs updated. Which is easy to forget, but it's also on the reviewers to remind of that, as you did many times before ^^ (and you should).

docs/todo.rst
```diff
@@ -1,8 +1,47 @@
...
+To check your updates, you can locally generate the documentation before pushing to master:
+
+.. code-block:: bash
+
+        $ cd docs/
+        $ make html
+
+Generale guidelines:
```
ivalaginja, line 40: "Generale" --> "General"

ivalaginja (commented):
Minor comments, otherwise ready to go.

johanmazoyer:
Ready for review

ivalaginja (approved):

PR #141, opened 2022-12-19
Write config copy to Labivew matrix directory

ivalaginja:
See title. It's an easy way to save all parameters with the matrices for now.

### DIFF
--- a/Asterix/main_THD.py
+++ b/Asterix/main_THD.py
@@ -143,6 +143,10 @@ def runthd2(parameter_file_path,
                           save_for_bench=onbench,
                           realtestbed_dir=labview_dir)
 
+    ### Write configfile to Labview-style matrix directory
+    config.filename = os.path.join(labview_dir, "Simulation_parameters.ini")
+    config.write()
+
     ### Set initial phase and amplitude
     # Phase upstream of the coronagraph (entrance pup)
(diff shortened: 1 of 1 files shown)


Asterix/main_THD.py
```diff
@@ -143,6 +143,10 @@ def runthd2(parameter_file_path,
                           save_for_bench=onbench,
                           realtestbed_dir=labview_dir)
 
+    ### Write configfile to Labview-style matrix directory
+    config.filename = os.path.join(labview_dir, "Simulation_parameters.ini")
```
johanmazoyer, line 147: Axel had dvp a tool to save all params to fits header also `from_param_to_header` in utils/save_and_read.py module. 
`header = from_param_to_header(config)`
This could be also interesting so that each configuration is attached to the matrix file itself

ivalaginja:
Just a ping for a review here please @johanmazoyer

Asterix/main_THD.py
ivalaginja: References in #152 - I would like to merge this here asap because we really need it when working on the testbed.

johanmazoyer: ok

johanmazoyer (approved):

Asterix/main_THD.py
johanmazoyer: ok

PR #142, opened 2022-12-20
Fix and upgrade regional sampling

johanmazoyer:
This PR is a fix of the regional sampling. Now can achieve 10-10 contrast with high enough parameters. 
- it corrects an error in the shift that was not uniformly done in all the dimension
- It does it better using the X_offset_output and X_offset_input parameters of the mft which allows shift without doing phase ramps
- It introduces value errors in a number of case and remove the outer_sampling parameter that was confusing
Currently include a lot of comments to document what I did, can be cleaned after
- in Asterix most pupil planes are over-padded but the current regional prop function assumed the pupil was exactly the size of the pupil plane array. Added a real_dim_input function, same properties as the one in mft. 
- updated tests

PR #145, opened 2022-12-21
Match WV phase mask orientation on THD2 testbed

ivalaginja:
See title.

I did this by comparing to the code setup of a commit that creates matrices that work on the testbed and manage to create a DH.

The phase masks look like this:
phase_masks
Left the old commit (213d669b) from fits file including a flip and a rotation, center the correctly rotated phase mask from Asterix to match that, here on this branch, right their difference normalized by the max of the fits-file version (left).

(Note how commit 08b91a is now [79fc551](https://github.com/johanmazoyer/Asterix/pull/145/commits/79fc551edcf252a4cbea6b5dc50753d8b6ae1008) because I rebased.)

This is completely unrelated to the type of propagation, we are going to test matrices generated with this but using MFTs on the hardware. If that works (🤞 ), we can focus on figuring out what's up with our zoom-in sampled propagations (see #142 and #144).

Below a comparison of interaction matrices from both commits. I take a random piece of the DM3 interaction matrix for Labview, as calcualted on commit 213d669b and calculated on this branch, subtract them, then normalize by the max of the commit 213d669b matrix. **Both matrices have been generated with MFT propagations.** Their relative difference is very small.
jacobians

### DIFF
--- a/Asterix/optics/coronagraph.py
+++ b/Asterix/optics/coronagraph.py
@@ -532,6 +532,9 @@ class Coronagraph(optsy.OpticalSystem):
                                                              inclination_x=inclination_x,
                                                              inclination_y=inclination_y)
 
+        # Match the orientation of the phase mask to THD2
+        phase_wrapped_vortex = np.rot90(phase_wrapped_vortex)
+
         wrapped_vortex = list()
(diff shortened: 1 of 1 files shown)


johanmazoyer (approved):
ok

Issue #147, opened 2023-01-03
Error in polychromatic ligtht


johanmazoyer:
@charles2719 found a quite important error.
In the mft, the change of wl is done by changing the number of resolution element in the exit image. Because the number of pixel stays the same, this correspond to a change in resolution. In monochromatic, this is define directly with the parameter `dimScience / Science_sampling` which are defined assuming the central wavelength `wavelength_0`. To scale to another wl, we use the ratio `lambda_ratio = wavelength / self.wavelength_0`. However,there is a mistake at several places where we multiply the  number of resolution elements instead of dividing. This is quite an important error and luckily catched in time before doing polychromatic intensively. 

The following code produce the following gif that can be tested at the right commit (before / after correction of this error) to see how the PSF scale with wl. The gif included show the current (before fixing the error) and show that the PSF decrease with WL ! 

```
import os
import numpy as np

import glob

from PIL import Image



from Asterix import Asterix_root
from Asterix.utils import read_parameter_file
from Asterix.optics import Pupil

from matplotlib.colors import LogNorm
import matplotlib.pyplot as plt

def make_gif(frame_folder):
    frames = [Image.open(image) for image in sorted(glob.glob(f"{frame_folder}/*.jpg"))]
    frame_one = frames[0]
    frame_one.save(f"{frame_folder}/my_awesome.gif", format="GIF", append_images=frames,
               save_all=True, duration=1000, loop=0)
    



dir_image = '/Users/jmazoyer/Desktop/gif_polychrom/'

parameter_file_test = os.path.join(Asterix_root, 'tests', "param_file_tests.ini")
config = read_parameter_file(parameter_file_test)

config["modelconfig"]['Delta_wav'] = 300e-9
config["modelconfig"]['nb_wav'] = 3

modelconfig = config["modelconfig"]

lapup = Pupil(modelconfig)

for i in range(config["modelconfig"]['nb_wav']):
    wavelength = lapup.wav_vec[i]

    psf = np.abs(lapup.todetector(wavelength = wavelength))**2


    fig = plt.figure()
    fig.set_size_inches(1,1, forward=False)
    ax = plt.Axes(fig, [0., 0., 1., 1.])
    ax.set_axis_off()
    fig.add_axes(ax)

    namejpg = f'PSF_wl{int(wavelength * 1e9)}.jpg'
    
[...]

PR #148, opened 2023-01-03
Fix error in lambda ratio

johanmazoyer:
Fix issue #147, problem in polychromatic light where the psf decrease with increasing wl instead of the opposite.


ivalaginja:
@johanmazoyer could you post a gif equivalent to that in #147 in here, with the fixed version?

johanmazoyer:
> @johanmazoyer could you post a gif equivalent to that in #147 in here, with the fixed version?

Here is the same gif as in #147 but with current branch :

my_awesome

ivalaginja (approved):
Thanks for catching the mistake @charles2719, thanks for fixing @johanmazoyer.

Issue #123, opened 2022-11-23
Move zoom-in sampled propagation of coronagraphs to `coronagraph.py`


ivalaginja:
> ok I see the problem. Even if you do that, I get a "circular import" warning when I do a "pylint Asterix/". this is not critical if we're careful I think, we've discuss that (hcipy have tons of these circular imports) but would be better if we can avoid that.
> 
> When you do  
`from Asterix.optics import`
you actually import all optics functions and you get lot of circular imports. What I meant by targeted is something like :
`import Asterix.optics.phase_amplitude_functions as phase_ampl`
you limit a lof the circular import because you only import `phase_amplitude_functions`.
> 
> However, this is a "true" circular import because `phase_amplitude_functions` need the mft, so you need to import `propagation_function`. It will probabaly not broke the stuff because you only need the mft in a `sine_cosine_basis` which is very specific, but I'm not sure in all cases.
> 
> Eventually the solution is to move `prop_fpm_regional_sampling` to coronagraph.py. I already asked you that, I think it make sense, because other propagation functions (MFT, FFT, Frensnel) in this file are very general, they are called by a lot of other optical instruments or even by other functions (like `sine_cosine_basis` or others in utils) because you always need fft-like functions. On the other hand `prop_fpm_regional_sampling` is only coronagraph oriented

_Originally posted by @johanmazoyer in https://github.com/johanmazoyer/Asterix/pull/119#discussion_r1030278827_

johanmazoyer:
ok

Issue #115, opened 2022-11-16
Increasing speed of MFT in Regional sampling 


johanmazoyer:
PR #80 increased the MFT by pre-calculating the matrices once and then only calculate the matrix multiplication during the coronagraph propagation. We should do the same for prop_fpm_regional_sampling.

The measure of the matrices is done at the end of the coronagraph initialization (around line 130-170 in optics/coronagraph.py). This can be done using mft with the option `returnAABB = True`.

Then during the propagation in `EF_through`, we use mft with  `only_mat_mult=True` and giving as parameters the AA and BB matrices calculated before (aroud line 380 in coronagraph.py).

PR #149, opened 2023-01-10
Move and clean regional sampling

johanmazoyer:
This PR moves the `prop_fpm_regional_sampling function` to coronagraph.py file to be more coherent (issue #123 ). We also sent the `butterworth_circle` function to optics/phase_amplitude_functions.py.

It introduce also a lot of small improvements to regional sampling: 
- clean the regional sampling in a single 'for' loop to have only 2 call to `mft` and not 6. 
- polychromatic wrapped vortex now working. After discussion with Raphael, we do not scale the regional sampling with wl because this method is only used with coronagraph that do not change radially. Only the phase mask is changed with wl.
- the number of pixel in the focal plane in regional sampling is now a specific parameter (`coronagraph.dim_fpm`) that is NOT the same one that the number of pixels in the final focal plane `coronagraph.dimScience`. The reason is that the first one is internal cooking and can be increased or decreased to go faster / obtain better contrast, whereas the latter is usually linked to the physical testbed / instrument and need to be fixed. 
- put all parameters for the regional sampling (FP pixel size and resolution of the different plane) in a single place at the beginning of the Coronagraph object initialization (arouns line 50 in optics/coronagraph.py)
- 'nbres' parameter is not optional parameter. The best numbers are heavily dependent on the size of the pupil and fpm in pixel, so we cannot just use a default value and hope for the best.

Change in the parameter file: 
- We changed the parameter "achrom_fqpm" to "achrom_phase_coro". This parameter apply to all coronagraphs that introduce phase (currently fqpm, vortex, wrapped vortex, and HLC) and allow us to choose whether we want the coronagraph to be achromatic (the coronagraph diplays the exact same phase shift for each wavelengths) or chromatic (phase is introduced by material step and therefore the introduced phase is different for each wl). 
- `Greypupil = True` is now the default in the example parameter file. 

Other small changes:
- added a print for the coronagraph attenuation **without** aberrations at the begining of each loop, just to check the numerical noise.
- a weird type error in the `prop_angular_spectrum`. I don't know why sometimes the number use to crop is recognized as an int and sometimes not. Now fixed always int

Finally, we modified the tests and website. 

Issues #89 (FQPM with regional sampling) and issue #115 (pre-calculate all mft matrices to go faster) are left for future PR but this is not a lot of work now.


Asterix/optics/coronagraph.py
```diff
@@ -817,3 +860,140 @@ def create_wrapped_vortex_mask(dim,
...
+    filter_order : int, default 15
+        Order of the Butterworth filter.
+    alpha : float, default 1.5
+        Scale factor for the filter size. The larger this number, the smaller the filter size with respect to the
+        input array.
+    dir_save_all_planes : string or None, default None
+        If not None, absolute directory to save all planes in fits for debugging purposes.
+        This can generate a lot of fits especially if in a loop, use with caution.
+
+
```
ivalaginja, line 913: One blank line too many.

Asterix/optics/coronagraph.py
```diff
@@ -817,3 +860,140 @@ def create_wrapped_vortex_mask(dim,
         phase_mask = phase
 
     return angles, phase_mask
+
+
+def prop_fpm_regional_sampling(pup,
```
ivalaginja, line 865: No action needed in this PR, but next time moving a block of code should not be done while also modifying that code... it is absolutely impossible for me to spot all the changes and my review is kinda useless here since I will miss potential issues. For next time.

Asterix/optics/phase_amplitude_functions.py
```diff
@@ -630,3 +630,57 @@ def make_sphere_lyot(dim_pp, prad, pupangle=0, spiders=True, grey_pup_bin_factor
                         reduce_outer_radius=lyotOuterEdgeObs,
                         add_central_obs=addCentralObs,
                         add_spider_thickness=addSpiderObs)
+
+
```
ivalaginja, line 634: One blank line too many.

Asterix/tests/test_coronagraph.py
```diff
@@ -14,11 +14,12 @@ def test_all_coronagraphs():
     # Reassign the parameter groups to variables
     modelconfig = config["modelconfig"]
     Coronaconfig = config["Coronaconfig"]
+    Coronaconfig.update({"filename_instr_apod": "RoundPup"})
 
     # Set coronagraph to be tested
     coros_to_test = ["fqpm", "wrapped_vortex", "classiclyot", "knife", "hlc", "vortex"]
-    expected_attenuation = [1e-20, 8e-10, 5e-3, 1e-2, 1e-3, 1e-20]  # Note that these are for the 100 px pupil below
-    atols = [0, 1e-14, 1e-14, np.nan, 1e-14, 0]  # zeros are for perfect coronagraphs
+    expected_attenuation = [1e-20, 5e-9, 5e-3, 1e-2, 1e-3, 8e-9]  # Note that these are for the 256 px pupil below
```
ivalaginja, line 21: Why did it get worse for the WV?

docs/run_asterix.rst
```diff
@@ -205,11 +205,12 @@ If HLC :
 
 If FQPM:
     - err_fqpm = 0 : float, phase error on the pi phase-shift (in rad)
-    - achrom_fqpm : bool, if True, Achromatic FQPM, else pi*lamda0/lamda
 
 If Vortex :
     - vortex_charge : even int, charge of the vortex
 
+If phase coronagraph:
+    - achrom_phase_coro = False : bool, If, we can choose to use it in achromatic mode (the coronagraph diplays the same behavior for each wavelengths) or not (phase is introduced by material step and therefore chromatic behavior)
```
ivalaginja, line 213: What's with the `If,`? ^^ Also,  I would replace "it" by "the phase coronagraph" to be explicit.

ivalaginja (changes requested):
Almost ready to merge except that I miss some sort of comparison of the new performance of the coro propagations vs. the old ones. Could you post some figures or tables (or both) in the PR that show the effective change in results you obtain with the changes in this PR?

Asterix/tests/test_coronagraph.py
johanmazoyer: Because I've greatly reduced the pixel size of the focal plane and pupil plane to go faster

- in current master : Size of the pupil is 256 before padding and 512 after padding, size of the FP (same as dimScience) is 400: 
Attenuation of 'wrapped_vortex is 5.357987020188405e-10 time for the corono propagation is 0.01989s
- with this PR Size of the pupil is 200 and 300 after padding, size of the FP is 256:
Attenuation of 'wrapped_vortex is 2.081871059763059e-09  time for the corono propagation is 0.00591s

Time is not done super properly because only for 1 corono propagation but gives an order or idea (faster by a facteur 2-3)

I decided that 2 10-9 is good enough for our purpose because this is the kind of contrast limit we want to get on the testbed. As you know, it's annoying to optimize because it is not super reliable, increasing the number of pixels sometimes get you worse contrast, or maybe you need to better optimize the number of element of resolution in the list each time you increase the number of pixel. I might MCMC the stuff and try to get the best contrast for both pupil plane and focal plane size under 256. Feel free to play with it if you have an insight on it.

Asterix/optics/coronagraph.py
```diff
@@ -51,6 +51,41 @@ def __init__(self, modelconfig, coroconfig, Model_local_dir=None):
...
+        elif self.corona_type in ("vortex", "wrapped_vortex"):
+            self.prop_apod2lyot = 'regional-sampling'
+            if self.prad < 100:
+                raise ValueError(f"In regional-sampling mode, 'diam_pup_in_pix' must be > 200 to be most accurate")
+
+            # with 'diam_pup_in_pix' = 256 these parameter give ~3 10-9 for the wrapped vortex
+            # it is possible to do better but would require higher pupil and fpm size
+            # Ideally try to find the 2 best 'nbres' so that we can get better than 10-10
+            self.dim_fpm = 256
+            self.nbrs_res_list = [4, 32]
```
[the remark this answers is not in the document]
johanmazoyer: yes, contrast is 4.3 10-9 for vortex charge 2 (see test_all_coronagraphs() function)

Asterix/Example_param_file.ini
```diff
@@ -55,7 +55,7 @@ onbench=True
     # if True, all the pupils will be created 10x larger than diam_pup_in_pix and 
     # then rebinned to diam_pup_in_pix at the exception of pupils that are read in 
     # a fit file, which are always rebinned from the sixe of the .fits
-    grey_pupils = False
+    grey_pupils = True
```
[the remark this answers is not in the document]
johanmazoyer: Well. FQPM is in perfect coronagraph currently so contrast is always 0.
If we remove perfect coronagraph we go from 6.98 10-6 (without grey pup) to 6.14 10-6 (with grey pup) with the parameters as they are in param_file_tests.ini

I agree it would probably be good practice to test on the tesbed also.

Asterix/optics/coronagraph.py
```diff
@@ -51,6 +51,41 @@ def __init__(self, modelconfig, coroconfig, Model_local_dir=None):
...
+        # define the types of propagation first for each coronagraph type
+        # this is also used to hard code some of the internal propagation of the
+        # coronagraph like the size of the focal plane
+        if self.corona_type in ("classiclyot", "hlc"):
+            self.prop_apod2lyot = 'mft-babinet'
+
+            self.rad_lyot_fpm = coroconfig["rad_lyot_fpm"]
+
+            # we oversample the center in babinet's mode because we can
+            # hard coded for now, this is very internal cooking
```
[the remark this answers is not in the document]
johanmazoyer: ok

Asterix/optics/coronagraph.py
```diff
@@ -51,6 +51,41 @@ def __init__(self, modelconfig, coroconfig, Model_local_dir=None):
         self.corona_type = coroconfig["corona_type"].lower()
         self.string_os += '_' + self.corona_type
 
+        # define the types of propagation first for each coronagraph type
```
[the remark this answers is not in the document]
johanmazoyer: the first is going through propagation methods and define the internal param (size of FP, etc) once for all coronagraph using this method
The second one is to define the FP once we have set the prop method and associated parameter.
This is the solution I found to group the internal param (size of FP, etc) once for all corono using this method.
I've commented better this part

Asterix/optics/phase_amplitude_functions.py
johanmazoyer: ok

docs/run_asterix.rst
johanmazoyer: done

Asterix/optics/coronagraph.py
```diff
@@ -51,6 +51,41 @@ def __init__(self, modelconfig, coroconfig, Model_local_dir=None):
...
+
+            # Following lines ~7 10-10 for 'diam_pup_in_pix' = 512. This is very slow though
+            # self.dim_fpm = 512
+            # self.nbrs_res_list = [6, 42]
+
+        elif self.corona_type in ("fqpm", "knife"):
+            self.prop_apod2lyot = 'mft'
+
+        else:
+            raise ValueError(f"The requested coronagraph mode '{self.corona_type}' does not exists.")
```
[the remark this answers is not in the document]
johanmazoyer: ok

Asterix/optics/coronagraph.py
johanmazoyer: ok

Asterix/Example_param_file.ini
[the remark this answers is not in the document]
ivalaginja: Oh right. Then it should be fine. Agreed with testing on the testbed, but we're not gonna change phase masks just for a software test and risk messing up our current alignment, so we'll manage as-is. Thanks for the clarification!

Asterix/tests/test_coronagraph.py
ivalaginja: Understood, thanks!

ivalaginja (commented):
Found one last typo, everything else was fixed and updated, or explained.

I still think it would still be good to create a table containing two or three pupil sizes in pixels (whichever you have tested), and resulting coronagraphic attenuation per coronagraph. This way we can come back to this PR and look it up for comparison without having to wade through the code. Would that be possible?

johanmazoyer:
> Found one last typo, everything else was fixed and updated, or explained.
> 
> I still think it would still be good to create a table containing two or three pupil sizes in pixels (whichever you have tested), and resulting coronagraphic attenuation per coronagraph. This way we can come back to this PR and look it up for comparison without having to wade through the code. Would that be possible?

I put array in the `prop_fpm_regional_sampling` docstring

ivalaginja:
Could I still motivate you to post it here too, please? Docstrings are much more dynamic and are likely to change further down the line. I would like to have a more permanent entry somewhere to really be able to pin it down to a point in time.
Also, looking at the table in the docstring tells me nothing about how it was tested - with pure arrays, like in the notebook, or with a full testbed object? If with a testbed object, with what configuration? That's too much information to put in a docstring.

I am approving this PR now since this discussion has nothing to do with the code.

ivalaginja (approved):

johanmazoyer:
> Could I still motivate you to post it here too, please? Docstrings are much more dynamic and are likely to change further down the line. I would like to have a more permanent entry somewhere to really be able to pin it down to a point in time. Also, looking at the table in the docstring tells me nothing about how it was tested - with pure arrays, like in the notebook, or with a full testbed object? If with a testbed object, with what configuration? That's too much information to put in a docstring.
> 
> I am approving this PR now since this discussion has nothing to do with the code.

The parameters of this function are very hard to optimize because it is not super reliable, increasing the number of pixels in PP or FP generally gives you better contrast but sometimes get you worse contrast. Following array gives you some of the result I obtained, using test_all_coronagraphs() in test_coronagraph.py with some those specific parameters:
    __________________________________________________________________________________________
    |     Dim pup    | Overpadding PP factor | Actual PP size | FP size |     nbres    || Resulting contrast |
    |       200         |                   1.5                |      300             |   256    |    [4, 32]   ||       2.1e-09            |
    |       280         |                   1.5                |      420             |   512     |    [6, 50]   ||       3.6e-10            |
   _________________________________________________________________________________________

PR #151, opened 2023-01-16
Optimize speed wrapped vortex / regional sampling

johanmazoyer:
This PR aims at speeding up the regional sampling for vortex. It does it with 2 improvements: 

- Pre-calculating the MFT matrix AA and BB matrix in both direction during the coronagraph initialization. We store the pre-calculated matrix and each MFT is only a (relatively small) matrix multiplication AA.image.BB (Issue #115)

- Found best parameters for wrapped vortex with regional sampling. The parameters of the `prop_fpm_regional_sampling` function are very hard to optimize because they're not super reliable. Increasing the number of pixels in PP or FP generally gives you better contrast but sometimes get you worse contrast. Following array gives you some of the result I obtained for the wrapped vortex:

Screenshot 2023-01-17 at 16 29 08

Method : I've tried to find the smallest pupil and focal plane pixel sizes for which we reach a target contrast (2.10-9, 10-9, 5.10-10, 2.10-10, 10-10) with only 3 samples zones in regional sampling (2 in the array + the outer one is always fixed to 2). These are probably the fastest possible way to simulate a wrapped vortex down to the desired contrasts. 

Careful, these numbers will not be optimized for classical vortex coronagraphs and not for fqpm although best parameters are probably close.


ivalaginja:
This helps and is good to merge after me, I would just like to have a master branch that works out of the box without raising an error. In this case, we need to merge either either 100 px in pupil + MFT, or 256 px in pupil + regional sampling.

Since the regional sampling is working now, how about we just set the pupil pixels to 256 in the config file, merge, I get the tagged release done and then we talk about how to continue with this @johanmazoyer ?

ivalaginja (approved):
Ready to go. We acknowledge that the 256 px in the pupil are overkill, but we're taking them because it's what's tested on hardware right now.

Merging this PR puts a halt on all other PRs into master until we figure out or hardware issues.

PR #159, opened 2023-02-14
Polychromatic Riemann Sum

johanmazoyer:
+ This PR introduces the Riemann Sum to do polychromatic images from monochromatic images not necessarily evenly sampled. It introduce a new parameter in the `[modelconfig]` section, `mandatory_wls`. This optional keyword allows the introduction of specific wavelengths that need to appear to simulate the polychromatic images. I've made it a "hidden parameter" which means it's not in the parameter files but if you know what you're doing you can modify this. 
These new wavelengths must be in the range ]wavelength_0 - Delta_wav / 2 , wavelength_0 + Delta_wav / 2[. Default is an empty list ('mandatory_wls =  , '). In this case, the nb_wav simulation wavelengths are chosen to be well equally distributed in the Delta_wav. Following figure is in the docs to understand both case: 

image

We then do a Riemann sum to create the polychromatic image:
image

+ This new option allows more freedom in the choice of the estimation and correction wavelength. This PR introduces a new parameter in `[Estimationconfig]`: `estimation_wls`. This allows us to hand pick the correction and estimation wavelengths that we want, even if they are not well centered or evenly space in the bandwidth. I have checked that photon noise should be working. 

+ I modified THD2 class to automatically include `estimation_wls` in `mandatory_wls`.

+ Lot of changes in the docs to be a better documentation

+ Solve some errors in polychromatic correction. This is now quite well tested and I think this is working and ready for the testbed

+ up to v2.6

Docs for this branch can be found here : https://asterix-hci.readthedocs.io/en/polychrom_rieman_sum/

Issue #138, opened 2022-12-15
Implement logging


ivalaginja:
It would help if the terminal output wasn't just written out to the console but also saved into a log file per experiment.

PR #160, opened 2023-02-21
Silence everywhere

johanmazoyer:
This PR extends the silence keyword everywhere. This is useful if we want to run tons of loops in a row to make a parameter analysis. Now silence mean silence. 

Please pull PR #159 before this one.


ivalaginja (approved):
Does what it's supposed to.

If you ever want to simplify this and make it easier to control, you should look into implementing a logger. That way you could get rid of passing the `silence` keyword around altogether (see also #138).

PR #164, opened 2023-04-14
Small changes suggested by pierre

johanmazoyer:
3 minor changes asked by Pierre:
- remove the "press any key to continue" in case you want to do a weird number of modes in the EFC inversion (very low for 2 DMs or the opposite). It still print a warning as it can be an easy way to break the correction and spend an afternoon without understanding why.
- change the PW matrix fits file. Because the exact wl of the laser can cahnge a bit but labview reads always the same file. Now the labview files are named "laser1", "laser2" and "laser3", instead of the exact wl of simulation => Associated Labview modification was performed by Raphael on 26/04/2023
- added an option to do a fits file DH mask

Another small change I did to call find_DM_to_probe function for the CDI workshop nb:
- remove find_DM_to_probe function from the estimation class so that it can be called from outside. This is just a change in the indent of the function basically


johanmazoyer:
SIMU 1dm HDH
hdh_1dm

SIMU 1DM FDH
fdh_1dm

SIMU 2DM FDH
Screenshot 2023-04-26 at 09 14 08

johanmazoyer:
THD 1dm HDH
johan - 20230426_HDH_1DM_laser3

THD 1dm FDH
johan - 20230426_FDH_1DM_laser3

THD 2dm FDH
johan - 20230426_FDH_2DM_laser3

johanmazoyer:
esa logo 2DM simu 

Screenshot 2023-04-26 at 10 37 03

johanmazoyer:
Simulation results : 
HDH 1DM - FDH 1DM
FDH 2DM
Screenshot 2023-04-26 at 15 01 18

johanmazoyer:
THD2 results : 

HDH - 1DM 
Screenshot 2023-04-26 at 16 50 19

FDH - 1DM
Screenshot 2023-04-26 at 16 51 37

FDH - 2DM
Screenshot 2023-04-26 at 16 55 12

ivalaginja:
@johanmazoyer would you mind adding a note about what coronagraph was used for the displayed figures? On the testbed it's pretty easy to see it was a FQPM, but in simulation it looks a but too clean to be able to recognize it immediately.

Asterix/Example_param_file.ini
```diff
@@ -184,9 +185,9 @@ onbench=True
     # lyot diameter (in meters). 
     # Only use in the case of a RoundPup Lyot stop.
     # not use for propagation. 
-    # Current value is 8.035mm = 8.1*0.97 
-    # (rayon Lyot * de-zoom entrance pupil plane / Lyopt plane)
-    diam_lyot_in_m  = 8.035e-3
+    # Current value is 7.984mm = 7.9*1.0107
```
ivalaginja, line 188: please make sure this also reads `7.895`, otherwise it gets confusing ^^

Asterix/Example_param_file.ini
```diff
@@ -278,7 +279,8 @@ onbench=True
     
     # Dark Hole Mask parameters
 
-    # "circle", "square", "noDH" or "custom"
+    # "circle", "square", "nodh" or "custom" (if custom, letters are harcoded in creatingMaskDH function in maskDH.py)
+    # Can also be a fits name with a mask of the right size (dimEstimxdimEstim)
```
ivalaginja, line 283: Please put spaces around the `x`, the current thing is fairly unreadable

Asterix/tests/param_file_tests.ini
```diff
@@ -26,6 +26,7 @@ onbench=True
 
     # Sampling in the detector science image
     #(lambda/Entrance_pupil_diameter in pixel
+    # For THD2 testbed : 7.42 pix / L/D for the 783.6 laser
```
ivalaginja, line 29: Repeating here because it is important:
Are you sure this really refers to the entrance pupil diameter? If this was measured on testbed data, the sampling of 7.42 pixels per lam/D is actually in Lyot stop diameter. We should absolutely clarify this. If this is indeed already converted to lam/D_pup, then we might at least subscript the D here to D_pup or something to make it very explicit.

Asterix/Example_param_file.ini
```diff
@@ -26,7 +26,8 @@ onbench=True
 
     # Sampling in the detector science image
     #(lambda/Entrance_pupil_diameter in pixel
-    Science_sampling = 7.25
+    # For THD2 testbed : 7.42 pix / L/D for the 783.6 laser
+    Science_sampling = 7.42
```
ivalaginja, line 30: Are you sure this really refers to the entrance pupil diameter? If this was measured on testbed data, the sampling of 7.42 pixels per lam/D is actually in Lyot stop diameter. We should absolutely clarify this.

Asterix/tests/param_file_tests.ini
```diff
@@ -184,9 +185,9 @@ onbench=True
     # lyot diameter (in meters). 
     # Only use in the case of a RoundPup Lyot stop.
     # not use for propagation. 
-    # Current value is 8.035mm = 8.1*0.97 
-    # (rayon Lyot * de-zoom entrance pupil plane / Lyopt plane)
-    diam_lyot_in_m  = 8.035e-3
+    # Current value is 7.984mm = 7.9*1.0107
```
ivalaginja, line 188: Same here, please make sure the number in the comment is the same like the number that is actually used.

Asterix/tests/param_file_tests.ini
```diff
@@ -278,7 +279,8 @@ onbench=True
     
     # Dark Hole Mask parameters
 
-    # "circle", "square", "noDH" or "custom"
+    # "circle", "square", "noDH" or "custom" (if custom, letters are harcoded in creatingMaskDH function in maskDH.py)
+    # Can also be a fits name with a mask of the right size (dimEstimxdimEstim)
```
ivalaginja, line 283: Please add spaces around the `x`

Asterix/tests/test_coronagraph.py
```diff
@@ -18,7 +18,7 @@ def test_all_coronagraphs():
 
     # Set coronagraph to be tested
     coros_to_test = ["fqpm", "wrapped_vortex", "classiclyot", "knife", "hlc", "vortex"]
-    expected_attenuation = [1e-20, 3e-9, 5e-3, 1e-2, 1e-3, 5e-9]  # Note that these are for the 200 px pupil
+    expected_attenuation = [1e-20, 1e-8, 5e-3, 1e-2, 1e-3, 5e-9]  # Note that these are for the 200 px pupil
```
ivalaginja, line 21: Where does this performance requirement decrease in this test come from? Did the test start failing with the changes in this PR?

Asterix/wfsc/estimator.py
```diff
@@ -209,7 +209,19 @@ def __init__(self,
...
+
+                    # Because the exact laser WLs can change a bit quite often and labview reads a given
+                    # fits name, we name the PW matrix laser1, 2, 3 for a range of WL.
+
+                    string_laser = ''
+                    if 625 < wave_k * 1e9 < 645:
+                        string_laser = "_laser1"  # ~635nm laser source
+                    if 695 < wave_k * 1e9 < 715:
+                        string_laser = "_laser2"  # ~705nm laser source
+                    if 775 < wave_k * 1e9 < 795:
```
ivalaginja, line 221: What happens if I want to run this at wvln = 750 nm in simulation?

ivalaginja (changes requested):

Asterix/Example_param_file.ini
johanmazoyer: done

Asterix/Example_param_file.ini
johanmazoyer: done

Asterix/tests/param_file_tests.ini
johanmazoyer: done

Asterix/tests/param_file_tests.ini
johanmazoyer: done

Asterix/tests/test_coronagraph.py
johanmazoyer: yeah, the regionnal sampling is so sensitive to these stuff that the change of Lyot size / FP resolution juste change the contrast by a factor of 4. I found a new point that seems to give 2.10-9 but this is just stupid to have that amount in the test because you cannot change your parameters.

Asterix/wfsc/estimator.py
johanmazoyer: This is not the simulatiom, this is the matrix taht are saved for the testbed in the labview directory and there are no laser at 750 on the tested. Pierre complained that because the code was saving with the actual wl in the simulation but labview was not updating matrix name. So if you change the simulation from 786 to 787, the matrix could not be read with the testbed. I just put some ranges so that it easier

Asterix/tests/param_file_tests.ini
johanmazoyer: I asked the same question to Pierre yesterday and he's affirmative

Asterix/tests/test_coronagraph.py
rgalicher: I don't understand. Do you mean the performance is very sensitive to one parameter for all coronagraphs? for one specific coronagraph (wrapped vortex)? Which parameter is it (size of the pupil in pixels?)?

johanmazoyer: This is only for the wrapped vortex. There are "hidden parameters" for the regional sampling: the size of the focal plane and the samplings of the regions (usually 2 different sampling + the outer one which is always nyquist in our case). for a given size of entrance pupil (for example 200x200), I found the smallest parameter that allow you to get under a certain limit in contrast (see array below)
Screenshot 2023-04-28 at 07 49 29

However, because I really tried to take them as small as possible (to go faster), they are very sensitive and can increase the simulated contrast if you change the type of corono (vortex or wrapped vortex) + size of Lyot + resolution in the focal plane, pupil size, etc. So we can either use a very oversized focal plane size (=1024), which will always give ok contrast but will be long or just work with higher contrast by default and optimize these number in specific case when you want to optimize your simulation. I don't think I should take time to optimize these number just for the testing each time

ivalaginja: I just wanted to clarify due to which parameter change the test contrast decreased, since not all that much changed in the relevant parameters. Is it due to the new Lyot stop diameter?

Asterix/wfsc/estimator.py
ivalaginja: Sure, but does this mean if I run this in simulation at one of the wavelengths not covered here, and forget to set `save_for_bench=False`, that it will raise an error? It might be worth catching that error and letting the code continue.

ivalaginja (approved):
Good to go for me, just asking for one clarification and one follow-up task.

PR #165, opened 2023-05-15
Small vortex upgrade

johanmazoyer:
Smalls changes to make Vortex easier to use on the testbed:
- Vortex is now use by default in MFT propagation + perfect corono option (at all wl). 
- Adapted the parameter files with more default option for vortex

### DIFF
--- a/Asterix/Example_param_file.ini
+++ b/Asterix/Example_param_file.ini
@@ -11,7 +11,7 @@ onbench=True
 [modelconfig]
 
     # central wavelength (in meters)
-    wavelength_0 = 783.6e-9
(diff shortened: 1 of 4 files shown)


ivalaginja (approved):
LGTM. I just added the sampling info to the other parameter files too.

ivalaginja:
I now pushed a new commit to delete them. I purposefully did not revert the commit so we keep the history of this weird thing.
Locally, the tests obviously pass since I am on a Mac and on here they only fail on Ubuntu.

ivalaginja:
Wow. Failing tests due to comments. Not going to deal with that now, I will just merge as is.

PR #167, opened 2023-11-17
Custom mask orientation 

johanmazoyer:
Add 2 lines in the mask creation to create a custom mask (e.g. CNRS) with the right orientation :
- For the correct orientation in simulation and in DS9 testbed data
- For the correct orientation on the testbed labview screen for demo

So that we do no have to redo all the time we want to do a demo


ivalaginja (approved):
LGTM

johanmazoyer:
This was tested for this purpose by me and raphael when we did it. I'm not redoing all the test because it's a very minor change to the custom mask so no change to the rest of the code bahvior

PR #169, opened 2023-12-07
update header matrices + saved thd-control version

johanmazoyer:
1) Lot of header change in most fits saved : 

a / Added the config is all saved matrices. Created specific fonctions to identify `name_header_efc_matrix` and `name_header_pwp_matrix` to specifically identify kw important for each matrix. Based on these kw, the matrix are recalculated or not (this is more selective than the name of matrix that was used until now)/ 

b/ Date and hour of saving is also saved in header["date_mat"] 

c/ Finally Added a kw : header["MODELSIM"] = 'asterix' in all matrices
Referenced in this issue in thd-controls package
https://github.com/ivalaginja/thd-controls/issues/132

3) we also save the matrices in the thd-controls format, see 
https://github.com/ivalaginja/thd-controls/issues/129

### DIFF
--- a/Asterix/wfsc/wf_control_functions.py
+++ b/Asterix/wfsc/wf_control_functions.py
@@ -280,18 +280,14 @@ def create_singlewl_interaction_matrix(testbed: Testbed,
         # For 1DM test / 2DM test
         # Matrix is saved/loaded for all the FP and then crop at the good size later
 
-        # we save the DM basis
-        list_str = basis_str.split('_')
-        list_str.pop(2)
-        name_basis_fits = "Basis" + '_'.join(list_str) + ".fits"
-        fits.writeto(os.path.join(matrix_dir, name_basis_fits), DM.basis, header, overwrite=True)
-
         if os.path.exists(os.path.join(matrix_dir, fileDirectMatrix + ".fits")) and (initial_DM_voltage == 0.).all():
             if not silence:
                 print("The matrix " + fileDirectMatrix + " already exists")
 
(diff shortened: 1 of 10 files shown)


Asterix/wfsc/wf_control_functions.py
```diff
@@ -600,8 +601,11 @@ def create_singlewl_interaction_matrix(testbed: Testbed,
                 plt.ioff()
             # We save the interaction matrix:
             if (initial_DM_voltage == 0.).all():
-                fits.writeto(os.path.join(matrix_dir, fileDirectMatrix + ".fits"),
-                             InterMat[:, pos_in_matrix:pos_in_matrix + DM.basis_size])
+                hdu_list = fits.HDUList([fits.PrimaryHDU(header=header)])
+                hdu_list.append(
+                    fits.ImageHDU(data=InterMat[:, pos_in_matrix:pos_in_matrix + DM.basis_size], name='MATRIX'))
```
ivalaginja, line 606: - Where does the transpose happen that is needed to feed this into `thd-controls`? Is this elsewhere? Could you point me to it please?

- We both agreed that "BOSTON" is not an ideal keyword value here, but to make it work with `thd-controls` for now, "MATRIX" needs to be replaced by "BOSTON". You could open an issue on `thd-controls` to discuss (and change it) later on, so that we don't forget about it.

johanmazoyer: This is internal to asterix here. The place where we save for thd controls is at the same place as we save for labview, in "corrector.py"

PR #184, opened 2025-04-15
Vectorize pw estimate calculation to improve speed

yaaaay:
Change function `calculate_pw_estimate` in `Asterix/wfsc/wf_sensing_functions.py` to use `np.einsum` 
instead of a nested `for` loop. The aim is to both clean up the code and improve the speed of the PW estimation.


johanmazoyer (approved):
Good to me

PR #179, opened 2024-11-13
Add THD2 optical elements as class attributes

ivalaginja:
This saves the already existing optical elements that build the THD2 optical system into actual class attributes so that they can be accessed from the outside.

Also adding a notebook that pulls the pupil array information out in a meaningful manner.

PR #188, opened 2025-07-08
Add notebook to create sinc probe files used with Roman HLC

ivalaginja:
Adds the full notebook in which the sinc-sinc-sine probes were created, inspected, and saved out to be used with the Asterix WFS&C loop.


johanmazoyer:
DM3 needs to be renamed DM2

ivalaginja:
Fixed, but needs https://github.com/johanmazoyer/Asterix/pull/179 merged first.

PR #198, opened 2026-06-08
Implemented perfect coronagrah

johanmazoyer:
This is the perfect coronagraph defined here : 
https://compass.pages.obspm.fr/website/tutorials/2023-07-19-coronagraph/

Careful, there was already something called "perfect coronagraph" in Asterix which is the technique where the Lyot aberrations are removed to make a "real coronagraph" (e.g. fqpm, vortex), perfect. This parameter was hardcoded so it should not interfere at the parameter file level. 

Tested that it's 0 in the absence of aberrations, at all wl.
Tested that it dig DH, slitlghy worse than "perfected FQPM" which is expected as there are more LO aberrations in case of perfect coronagraph

### DIFF
--- a/Asterix/optics/coronagraph.py
+++ b/Asterix/optics/coronagraph.py
@@ -52,7 +52,7 @@ class Coronagraph(optsy.OpticalSystem):
 
             self.string_os += '_' + "perfcoro"
 
-            self.pup_for_perfect_coro = pupil.Pupil(modelconfig, prad=self.prad)
+            self.pup_for_perfect_coro = pupil.Pupil(modelconfig, prad=self.prad, PupType=coroconfig["filename_instr_pup"])
 
             # Initialize the max and sum of PSFs for the normalization to contrast
             self.measure_normalization()
@@ -253,7 +253,7 @@ class Coronagraph(optsy.OpticalSystem):
                 # of the coronagraph to a round pupil to remove it
                 # THIS IS NOT THE ENTRANCE PUPIL,
                 # this is a round pupil of the same size
-                pup_for_perfect_coro = pupil.Pupil(modelconfig, prad=self.prad, PupType=coroconfig["filename_instr_pup"])
+                pup_for_perfect_coro = pupil.Pupil(modelconfig, prad=self.prad)
 
                 # do a propagation once with self.perfect_Lyot_pupil = 0 to
                 # measure the Lyot pupil that will be removed after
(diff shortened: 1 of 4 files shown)


Asterix/optics/coronagraph.py
```diff
@@ -40,6 +40,26 @@ def __init__(self, modelconfig, coroconfig, Model_local_dir=None, silence=False)
...
 
+        self.corona_type = coroconfig["corona_type"].lower()
+
+        # Special case of the perfect coronagraph
+        # (defined here : https://compass.pages.obspm.fr/website/tutorials/2023-07-19-coronagraph/)
+        # we skip all definition as this coronagraph doe not have the normal attribute of a coronagraph
+        # (no apodizer, no focal plane mask, no Lyot stop).
+        if self.corona_type == "perfect":
+            if modelconfig["filename_instr_pup"] != "RoundPup":
+                raise ValueError("For a perfect coronagraph, the Pupil must be round: set modelconfig['filename_instr_pup'] to 'RoundPup'.")
```
rgalicher: We can define a perfect coronagraph whatever the pupil.
Electric field E in the entrance pupil plane : E = P exp(i phi) with P the binary pupil (that can be whatever shape) and phi the phase aberration (can be complex if transmission aberrations exist).
Electric field E_2 in a pupil plane after the perfect corongraph : E_2 = E - <E> P where <E> is the spatial average of E over the non-zero pixels of P (i.e. where flux is transmitted through P).

Asterix/tests/test_coronagraph.py
```diff
@@ -17,12 +17,16 @@ def test_all_coronagraphs():
...
+    coros_to_test = ["fqpm", "wrapped_vortex", "classiclyot", "knife", "hlc", "vortex", "perfect"]
+    expected_attenuation = [1e-20, 1e-8, 5e-3, 1e-2, 1e-3, 5e-9,  1e-20]  # Note that these are for the 200 px pupil
+    atols = [0, 1e-14, 1e-14, np.nan, 1e-14, 1e-14, 0]  # zeros are for perfect coronagraphs
 
     for i, coro in enumerate(coros_to_test):
         Coronaconfig.update({"corona_type": coro})
+        if coro == "perfect":
+            # for the last coronagraph of the list, we modify the modelconfig to be compatible
+            # with a perfect coronagraph, which requires a round pupil and no grey pupils.
+            modelconfig.update({"filename_instr_pup": "RoundPup", "grey_pupils": False})
```
rgalicher: to be changed if non-round pupil are accepted for the perfect coronagraph.

johanmazoyer: done

Asterix/optics/coronagraph.py
johanmazoyer: done

rgalicher (dismissed):

PR #201, opened 2026-07-27
Allow loading of initial dm maps in the parameter file

johanmazoyer:
- add DM1_filename_initcommand and DM2_filename_initcommand to load initial DM command
- commands are used as basis when we create the EFC matrix
- When saving the loop result, I corrected a small error in the saving of all the loop DM commands that was not done properly. I also added another file "DM*_last_command.fits" that only contains the last DM command, so that can it be used directly in the parameter file if we want to restart the correction were it was stopped.

### DIFF
--- a/Asterix/main_THD.py
+++ b/Asterix/main_THD.py
@@ -146,7 +146,7 @@ def runthd2(parameter_file_path,
     for DM_name in thd2.name_of_DMs:
         if thd2.config_file["DMconfig"][DM_name + "_filename_initcommand"] != "":
             thisDM_initcommand = fits.getdata(config["DMconfig"][DM_name + "_filename_initcommand"])
-            DMs_initcommand += thd2.testbed_command_to_indiv_DM_command(thisDM_initcommand, DM_name)
+            DMs_initcommand += thd2.indiv_DM_command_to_testbed_command(thisDM_initcommand, DM_name)
 
     # Initialize the estimation
     estimator = Estimator(Estimationconfig,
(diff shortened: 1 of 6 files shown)


rgalicher (dismissed):

Asterix/main_THD.py
```diff
@@ -140,6 +141,13 @@ def runthd2(parameter_file_path,
     # Concatenate into the full testbed optical system
     thd2 = THD2(config, model_local_dir, silence=silence)
 
+    # Read if there was DM initial command
+    DMs_initcommand = np.zeros(thd2.number_act)
+    for DM_name in thd2.name_of_DMs:
+        if thd2.config_file["DMconfig"][DM_name + "_filename_initcommand"] != "":
+            thisDM_initcommand = fits.getdata(config["DMconfig"][DM_name + "_filename_initcommand"])
+            DMs_initcommand += thd2.indiv_DM_command_to_testbed_command(thisDM_initcommand, DM_name)
```
rgalicher, line 149: cette fonction crée bien un vecteur avec la bonne dimension et des 0 pour le DM qui n'est pas DM_name ?

johanmazoyer: oui :
> Returns
>         --------
>         testbed_command : 1D-array real of dim testbed.number_act
>             the commands of the testbed DM with dmcommand_indiv at the
>             position of DM DM_name and zero elsewhere

du coup en parcourant tous les DMs, on fait une commande "testbed" de taille 1024+954

rgalicher: c'est bien fait ! :-)

rgalicher (approved):

